PR #124, opened 2019-09-14
WIP - Xiaomi not showing added views

mzorz:
**NOTE: THIS IS WIP**

This fixes a weird bug that only appeared on a Xiaomi  Mi 8 with Android 8.1 phone, where adding emojis would just not appear, even though the views were added to the hierarchy these would simply not be displayed.
The views would immediately appear once the TextEditorDialog got displayed, which was weird. At first I thought the Dialog was doing something, and thought of the insets and window handling we're doing to keep things full screen, but figured out it wasn't the case.
I found the same problem in an issue in the React Native repo here https://github.com/facebook/react-native/issues/17968 (look into the “Actual Behavior” subtitle in the description). A few comments down there’s a workaround using Choreographer. Been reading a bit about [Choreographer](https://developer.android.com/reference/android/view/Choreographer) and how views are measured and how it plays a role with updating the hardware display, and it sounded like worth trying.

Right now it misplaces the emoji at top/left corner of the screen, and the "delete" area button doesn't appear (haven't figured out why yet).

IMG_20190914_104428

Note: if you try this code on a Pixel 2, there are noticeable artifacts such as the emoji will flicker and change position very quickly so, still need to make sure and work more on it.

### DIFF
--- a/app/src/main/java/com/automattic/portkey/compose/ComposeLoopFrameActivity.kt
+++ b/app/src/main/java/com/automattic/portkey/compose/ComposeLoopFrameActivity.kt
@@ -471,13 +471,6 @@ class ComposeLoopFrameActivity : AppCompatActivity() {
         )
     }
 
-    private fun testEmoji() {
-        val emojisList = PhotoEditor.getEmojis(this)
-        // get some random emoji
-        val randomEmojiPos = (0..emojisList.size).shuffled().first()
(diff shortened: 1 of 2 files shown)


mzorz:
We've got reports in alpha1 that this was working on the same Xiaomi Mi 8 device so, closing now

PR #278, opened 2020-03-15
Introducing Frame Saving process into a Service

### LINKED ISSUE #273: Move save frames process to Service
After implementing #257, it can take some time for a Story to get fully saved, especially if it contains several frames and even more if some of the frames use video as a background.

We should move processing to a Service, and trigger the process from the Activity's ViewModel.

I was thinking of directly implementing it on WPAndroid given we need to integrate the functionality there, but since a Service is part of the Application (and needs be declared in the Manifest), we'll make sure to have it implemented in the Portkey app before moving it to WPAndroid, in order to make the functionality reusable elsewhere as well, by using Portkey as a base model.

mzorz:
Closes #273 

### Intro
This PR takes the former work done in order to implement a Story Frames Save architecture (see #257 and related PRs) and encapsulate the work into a Service.

This is needed given sometimes, saving a Story's frames may take time (especially if they have video-background in one or more of their Story frames), and we don't want to block the user on a screen, waiting for things to get processed.

### Services
The Android documentation describes a few of the possibilities we have for performing [background work](https://developer.android.com/guide/background). We're going to use a Foreground Service given that's the mechanism that best fits the description of our use case as explained here.

> For user-initiated work that need to run immediately and must execute to completion, use a foreground service. Using a foreground service tells the system that the app is doing something important and it shouldn’t be killed. Foreground services are visible to users via a non-dismissible notification in the notification tray.

In our case, the user taps on `Save` button and makes it explicit that they want their Story and Story frames to get saved. Hence, it makes sense to encapsulate this work in a Foreground Service.

In our specific case, we can roughly say there are 2 big parts to finishing a Story:
1. Saving the composed Story frames to disk so these are available for upload
2. upload the frames and Story to a server.

For now, this PR will start with part 1, and move the already existing code from being carelessly called from the Activity (in the click listener for the `Next` button) into a foreground Service, as this operation can potentially be lengthy depending on hardware and amount of frames to be saved. 
For the upload part we're going to delegate that to the existing WordPress Android `UploadService`, which will manage both the adding of images and uploading a custom typed Post to an existing site (to be covered in further PRs when integration of the functionality into WPAndroid is planned).

### How it works
There's a problem I found while doing this: a [started Service](https://developer.android.com/guide/components/services#CreatingAService) is passed parameters with an Intent extras bundle, and our parameters here would be a list of images with their added Views, so unless we access the `ViewModel`'s `Repository` from the `Service` (with the Repository actually persisting the data, which is not our case right now as we just keep these in memory) then we can't pass the frames with their added views if we don't first serialize this information (that is, positioning, rotation and scale information for each of the added views). 

Given our Service is actually triggered from an Activity and is quite related to it (it's a user initiated action), looks like this is a good candidate for a [bound service](https://developer.android.com/guide/components/bound-services#Creating), thus allowing us to bypass the serialization process which among other things would also incur into some time penalty (both for execution and development time). In this case, for the reasons outlined above we also need to making it both a bound Service _and_ a foreground service (as we need the background saving process to exist even if the user navigates away from the Activity or even further, away from the app itself). This PR only tackled the bound Service part, and leaves the foreground Service for an upcoming PR so to keep things simpler and easier to review.

Once the processing is ready, it can either have a successful ending, or some of the frames may have not been saved successfully. The error collection strategy will be tackled on another PR, for now we're introducing EventBus here just as a means to signal the saving process has ended (either successfully or not, it doesn't matter for now in this PR), so the event listener can properly do the cleanup they want.

### Next steps (to be tackled in subsequent PRs):
- [x] implement foreground service (with foreground notification) https://github.com/Automattic/portkey-android/pull/279
- [ ] implement error collection strategy
- [ ] implement error collection UI

### A word on future WPAndroid integration
Before tackling this PR, I studied how other services were implemented, particularly [SiteCreationService](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/services/SiteCreationService.kt) as a model for a well written Service using the latest guidelines and good stuff Kotlin brings (such as corroutines) and the [UploadService](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java) given our process output here will be fed into the UploadService afterwards. We followed the same pattern as used for other Services in WPAndroid, which will make it easier later when we integrate this functionality there:
- the new class extends Service
- we make it a foreground Service, like the UploadService is. Although a Story can be saved in just a few seconds (while this depends, it's generally less than 2 seconds for ~14 frames as per our tests), we could also improve this part later when integrating with WPAndroid and make use of `AutoForeground` which is implemented there for the `SiteCreationService`.

#### To test:
1. open portkey
2. tap on the + FAB
3. either capture pictures or hold the big button on the bottom to start recording a video, several times
4. add some text / emoji to your frames (this makes processing harder, just 1 added view should be fine for this PR purposes)
5. once you're ready, tap NEXT
6. observe a "Saving..." scrim appears (navigating away from this will probably break stuff as this is just introducing a bound Service, so please just stay there for now :) )
7. once the process ends, the "Saving" scrim disappears
8. got to the Photos app and find your newly created frames (videos, pictures) with your added text / emoji there.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -55,6 +55,8 @@ dependencies {
     implementation 'com.github.bumptech.glide:glide:4.9.0'
     kapt 'com.github.bumptech.glide:compiler:4.9.0'
 
+    implementation 'org.greenrobot:eventbus:3.1.1'
+
     implementation 'io.sentry:sentry-android:1.7.27'
     // this dependency is not required if you are already using your own
     // slf4j implementation
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,6 +59,11 @@
             android:theme="@style/AppTheme.NoActionBar"
             />
 
+        <!-- Services -->
+        <service
+            android:name=".compose.frame.FrameSaveService"
+            android:label="Frame Save Service" />
+
         <!-- FileProvider used to share photos with other apps -->
         <provider
             android:name="androidx.core.content.FileProvider"
--- a/app/src/main/java/com/automattic/portkey/compose/ComposeLoopFrameActivity.kt
+++ b/app/src/main/java/com/automattic/portkey/compose/ComposeLoopFrameActivity.kt
@@ -82,9 +82,6 @@ import com.bumptech.glide.load.resource.bitmap.RoundedCorners
 import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.activity_composer.*
 import kotlinx.android.synthetic.main.content_composer.*
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
 import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
--- a/app/src/main/java/com/automattic/portkey/compose/ComposeLoopFrameActivity.kt
+++ b/app/src/main/java/com/automattic/portkey/compose/ComposeLoopFrameActivity.kt
@@ -129,7 +129,6 @@ class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelectorTapped
     private var isEditingText: Boolean = false
 
     private lateinit var storyViewModel: StoryViewModel
-    private lateinit var frameSaveManager: FrameSaveManager
     private lateinit var transition: LayoutTransition
 
     private lateinit var frameSaveService: FrameSaveService
@@ -140,7 +139,10 @@ class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelectorTapped
             Log.d("ComposeLoopFrame", "onServiceConnected()")
             val binder = service as FrameSaveService.FrameSaveServiceBinder
             frameSaveService = binder.getService()
-            frameSaveService.saveStoryFrames(0, frameSaveManager, StoryRepository.getImmutableCurrentStoryFrames())
+            frameSaveService.saveStoryFrames(0,
+                FrameSaveManager(photoEditor),
+                StoryRepository.getImmutableCurrentStoryFrames()
+            )
             saveServiceBound = true
         }
 
@@ -230,8 +232,6 @@ class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelectorTapped
                 delete_view.setReadyForDelete(ready)
             }
         })
-
-        frameSaveManager = FrameSaveManager(photoEditor)
 
         backgroundSurfaceManager = BackgroundSurfaceManager(
             savedInstanceState,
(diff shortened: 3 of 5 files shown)


jd-alexander (changes requested):
Thanks for these changes @mzorz I tested this and it worked! 
I even exited the app and the service continued in the background and I was able to see the newly created stories in my gallery. 

I stumbled on two crashes that you could resolve before we proceed. 

```
 Process: com.automattic.portkey, PID: 28239
    java.lang.IllegalStateException: Fragment already added: EmojiPickerFragment{c8ed3ee (de450151-c4e1-4835-a2c7-c00ac471d1cb)}
        at androidx.fragment.app.FragmentManagerImpl.addFragment(FragmentManagerImpl.java:1379)
        at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:399)
        at androidx.fragment.app.FragmentManagerImpl.executeOps(FragmentManagerImpl.java:2079)
        at androidx.fragment.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:1869)
        at androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManagerImpl.java:1824)
        at androidx.fragment.app.FragmentManagerImpl.execPendingActions(FragmentManagerImpl.java:1727)
        at androidx.fragment.app.FragmentManagerImpl$2.run(FragmentManagerImpl.java:150)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

```
    java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at java.util.ArrayList.get(ArrayList.java:437)
        at java.util.Collections$UnmodifiableList.get(Collections.java:1356)
        at com.automattic.portkey.compose.story.StoryViewModel.getCurrentStoryFrameAt(StoryViewModel.kt:84)
        at com.automattic.portkey.compose.ComposeLoopFrameActivity.addCurrentViewsToFrameAtIndex(ComposeLoopFrameActivity.kt:598)
        at com.automattic.portkey.compose.ComposeLoopFrameActivity.access$addCurrentViewsToFrameAtIndex(ComposeLoopFrameActivity.kt:105)
        at com.automattic.portkey.compose.ComposeLoopFrameActivity$addClickListeners$9.onClick(ComposeLoopFrameActivity.kt:542)
        at android.view.View.performClick(View.java:7125)
        at android.view.View.performClickInternal(View.java:7102)
        at android.view.View.access$3500(View.java:801)
        at android.view.View$PerformClick.run(View.java:27336)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

jd-alexander (approved):
> Let me know if there are any other observations that I'm happy to look into @jd-alexander !

Okay great! Yes, once there aren't related to this PR and their are issues to track them then we are all good here and ready to go!

LGTM 🚢

PR #298, opened 2020-04-10
Save frame Service primary PR

mzorz:
Contains all of the work related to containing the save frame operation in a Service and includes snackbars, error dialogs, error handling, and notifications.

Also now contains all of the alpha1 work and fixes


mzorz:
Merging 🚂

PR #464, opened 2020-07-27
Normalize exported images to 9:16

### LINKED ISSUE #451: Always export 9:16 media
From @megsfulton 👍 

To reduce the amount of variability with hiding parts of the media, one thing we might want to consider is having the mobile editors always output a media size that is 9:16, regardless of the device screen ratio.

Here’s how that might look. I don’t believe it would require any UI changes if we crop from the bottom “safe area” where the slide thumbnails are.

always-9_16-output-android-editing

#### Follow up notes / questions on this from  me:

 [can you confirm] (the bottom part of the screen is what gets cropped, variably depending on the device screen size and adjustments to get a 9:16 out of it).

> MF: That’s correct. To make it extra obvious to the user, we could make the opacity on the bottom bar 100% black on devices taller that 9:16

screen-shot-2020-07-14-at-9 35 35-am

#### Further information gathered

Most of the formats we can hope to be able to produce are 9:16 but there will still be cases where we can only produce 240×320 (4:3) or 480×720 (3:2).
In more detail: the new produced video will be of a different size than the original and this time it’ll be portrait oriented, to include the added panda emoji that sits on the now otherwise blank area in the example. The output video format (including the video frame size) will depend on what the device can do for us (also may or may not be the quality of the original video, for example), and is going to be normalized to these sizes:

```
// MUST
// - 320 x 240 px
// - 720 x 480 px
// SHOULD (when hardware available)
// - 1280 x 720 px
// - 1920 x 1080 px
```

In addition to these 4 up there, we also have these other 2 formats for bigger screens as we saw these are also widely supported nowadays on more recent Android versions:

- 1440p = 2560×1440
- 2160p = 3840×2160

While most of them are 9:16, the only 2 that are marked MUST are not which are 480p (3:2) and 240×320 (4:3)

mzorz:
Fix #451 

Title has it, in this PR we added method `normalizeHeightExportTo916()` to ensure the `clonedViewSpecs` adjust the measures to 9:16.

This is how it works:
1. if the screen is 16:9, we're OK and return as usual
2. if not, we resolve what the height should be for the screen's given width.
3. if the result of (2) is less than actual height, set the new target height for the cloned view to the number calculated as of (2) (cropping)
4. if the result of (2) is greater than actual height --> crop the width to match the ratio needed for the given height

Step (4) here is less (never?) likely to happen in practice, given most popular screens are some sort of not-square device where the width is always less than height (portrait mode), but it's coded so we make sure we have a defined state for all cases.

For example, using a Pixel 3 with a screen of 1440x2792:

**Before**:
produces an image of 1440x2792 (matching the screen's resolution)
https://testmzorz3.files.wordpress.com/2020/07/wp-1595874869618.jpg

**After**:
produces an image that is now 1440x2560 (matching the target ratio)
https://testmzorz3.files.wordpress.com/2020/07/wp-1595874869642.jpg

Will continue testing and add here.

To test:
1. take a phone that is not 16:9
2. install portkey and produce an slide with an image background
3. verify the output image is 16:9


mzorz:
Closing in favor of #479

Issue #465, opened 2020-07-28
Text Editor Dialog: Enhanced color picker


aforcier:
Replace current color picker row with button which brings up a bottom sheet. This also ties in with #170.

![text-editing-colors-1](https://user-images.githubusercontent.com/9613966/88614197-30dd4480-d0ca-11ea-8e72-7c39f5a73684.png)

This will also address the need for a chosen color indicator, so this issue supersedes #154.

PR #466, opened 2020-07-28
Text Editor: Support text alignment

### LINKED ISSUE #186: Edit screen: Support text alignment

aforcier:
Closes #186. Adds a new text alignment button to the text editor dialog which cycles text alignment through left, center, and right-aligned:

wp-stories-text-alignment

I also made a few changes to the text editor layout, using a placeholder icon for the color picker (to be used properly with #465), and for now making it toggle the current color picker view so it doesn't overlap the alignment button.

I'm targeting the `feature/text-editing-v2` branch for now, which is branched off the alpha2 branch - once we've merged alpha 2 and shipped a new alpha this can be merged all the way to `develop`, but in the meantime I'll keep the text editing v2 work in the feature branch.

### Notes

I had to use the `gravity` property to align the EditText inside the editor, but `textAlignment` for the resulting TextView added to photoeditor (noted in a comment [here](https://github.com/Automattic/portkey-android/pull/466/files#diff-6e706ba2e92e5195fa534191694c2501R110)). The gravity handling is encapsulated, so outside the editor (e.g. for serialization) we're only talking about `textAlignment`. But still a bit confusing, I'm happy to try something else if you have any ideas @mzorz .

### To test

1. Add multi-line text to the editor, toggle the alignment with the button
2. Add the view to the canvas, make sure the alignment value is respected
3. Re-edit an existing text, change the alignment, save again, make sure all is persisted properly
4. Turn on Don't Keep Activities, verify that text alignment value is correctly stored and de-serialized


photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
```diff
@@ -236,6 +236,7 @@ class PhotoEditor private constructor(builder: Builder) :
     fun addText(
         text: String,
         colorCodeTextView: Int,
+        textAlignment: Int? = null,
```
mzorz, line 239: wondering if we should make this non-nullable, and make the  default for it be `start` or `end` depending on the [Rtl setting on the device ](https://developer.android.com/reference/android/text/TextUtils#getLayoutDirectionFromLocale(java.util.Locale))🤔 (we could make it `start` for now)

mzorz (approved):
This is really nice to see working @aforcier 👍 🎉 
I've been looking into the alignment/gravity issue you mentioned, and it seems to be a cause of problems elsewhere. Gravity has been added later and seems preferable to use perhaps? 
This is indeed not as clear as one would ideally wish, but it does the work and the comments in code clarify it enough for me 👍 
Also, TIL `textAlignment` is a property of `View`, and not of `TextView` as I'd have imagined.

Left a minor comment, feel free to ignore it.

Approving 👍 , leaving the merge to you in case you need to check something with your base branch

photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
aforcier: Hmm 🤔 further down we're only using it if it's not null, in that case the view should just be initialized with the default text alignment which should automatically respect the RtL setting on the device. Same treatment we're doing for typeface, I think this is okay. But I may have missed part of your reasoning, let me know if that's the case.

aforcier:
Thanks @mzorz! I updated the base branch to the rename one from #470 since there was going to be a conflict there, and merged it in and resolved the conflict. We can merge this after #470 is in. I'll drop the `feature/text-editing-v2` branch I was using. future text editor PRs can just target `develop` directly at this point.

photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
mzorz: > in that case the view should just be initialized with the default text alignment which should automatically respect the RtL setting on the device

Ah that's fine then :+1:

PR #470, opened 2020-07-29
Update naming to stories

aforcier:
Use stories consistently throughout the codebase, and also rename the base app (at the moment effectively a demo app) to Loop.

Counterpart WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/12546

To test:
* Make sure everything runs
* I made some changes to the tooling store to reflecting the new naming, so confirm that CI is passing and an APK is generated

PR #471, opened 2020-07-29
Remove kotlin.code.style from example properties

aforcier:
We're using our own synced rules, and having this in the real gradle.properties keeps changing the rules and generating changes to tracked .idea files.

I removed it from my own gradle.properties a while ago but thought we should sync it centrally.

Relies on #470 since that one has changes to `gradle.properties-example` as well.


mzorz (approved):
👍 :shipit:

PR #478, opened 2020-08-04
Stories projects interoperability: strip Android specifics

mzorz:
See #477 first

### Discussion

Here I'm proposing some of the changes to the current serialization format, to get to a format that works for both Android and iOS.

- 28b78d27831981cc1e2df2cc539d871f6455f59a removed SaveResultReason which was a very local state representation
- 7188bcdd2c2532a09e6956464bc5960599f438d6 normalized `fontSize`, `textColor`, `textAlignment`
- 186679d54481c579e56fb48418cf2b5f5621fedb added `metadata` object which tells more about the context in which this Story has been created:

```
  "metadata": {
    "revision": "1.0",
    "device": {
      "screen": {
        "width": 1080,
        "height": 1280
      },
      "platform": "android|ios",
      "model": "iphone10"
    }
  },
```

- the `revision` is the interop JSON definition revision number to correctly attempt deserialization.
- the `device` object for now contains a screen width/height measures as informed by the device, given all other positioning measures will be relative to this canvas size.
- `platform` and `model` are self-explanative. We can add `brand` or `vendor` too which makes sense for Android.

Adding more comments on the code itself down here in the PR 👍

### DIFF
--- a/stories-interop-example.json
+++ b/stories-interop-example.json
@@ -1,4 +1,15 @@
 {
+  "metadata": {
+    "revision": "1.0",
+    "device": {
+      "screen": {
+        "width": 1080,
+        "height": 1280
+      },
+      "platform": "android|ios",
+      "model": "iphone10"
+    }
+  },
   "frames": [
     {
       "source": {
--- a/stories-interop-example.json
+++ b/stories-interop-example.json
@@ -18,9 +18,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "this is some text",
-              "fontSizePx": 140,
-              "textColor": -1,
-              "textAlignment": 2
+              "fontSize": 14,
+              "textColor": "#00000000",
+              "textAlignment": "left|center|right"
             }
           },
           "uri": null
@@ -46,9 +46,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "this is some text, on a video background",
-              "fontSizePx": 140,
-              "textColor": -1,
-              "textAlignment": 4
+              "fontSize": 14,
+              "textColor": "#ffffffff",
+              "textAlignment": "left|center|right"
             }
           },
           "uri": null
@@ -62,9 +62,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "😁",
-              "fontSizePx": 284,
-              "textColor": -16777216,
-              "textAlignment": 4
+              "fontSize": 28,
+              "textColor": "#ffffffff",
+              "textAlignment": "left|center|right"
             }
           },
           "uri": null
@@ -89,9 +89,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "one text view",
-              "fontSizePx": 140,
-              "textColor": -1,
-              "textAlignment": 2
+              "fontSize": 14,
+              "textColor": "#ffffffff",
+              "textAlignment": "center"
             }
           },
           "uri": null
@@ -105,9 +105,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "another text view",
-              "fontSizePx": 140,
-              "textColor": -1,
-              "textAlignment": 2
+              "fontSize": 14,
+              "textColor": "#ffffffff",
+              "textAlignment": "center"
             }
           },
           "uri": null
@@ -121,9 +121,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "third text view",
-              "fontSizePx": 140,
-              "textColor": -1,
-              "textAlignment": 2
+              "fontSize": 14,
+              "textColor": "#ffffffff",
+              "textAlignment": "center"
             }
           },
           "uri": null
@@ -137,9 +137,9 @@
             "scale": 1,
             "addedViewTextInfo": {
               "text": "😬",
-              "fontSizePx": 284,
-              "textColor": -16777216,
-              "textAlignment": 4
+              "fontSize": 28,
+              "textColor": "#ff55ff",
+              "textAlignment": "right"
             }
           },
           "uri": null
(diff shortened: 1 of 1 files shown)


stories-interop-example.json
```diff
@@ -1,4 +1,15 @@
...
+        "width": 1080,
+        "height": 1280
+      },
+      "platform": "android|ios",
+      "model": "iphone10"
+    }
+  },
   "frames": [
     {
       "source": {
```
mzorz, line 15: regarding this:
         `"type": "com.wordpress.stories.compose.story.StoryFrameItem.BackgroundSource.FileBackgroundSource",` 
I'm not sure we can change this on Android - we could by writing our own serializer for this class though, just commenting here. Maybe we can keep the format, even if this verbose?

stories-interop-example.json
```diff
@@ -18,17 +29,14 @@
             "scale": 1,
```
mzorz, line 29:
```
             "translationX": -180,
            "translationY": -861.4966,
```
these would remain the same, and are always calculated from the center of the screen where the object is initially placed. Added the screen size for this as part of the `metadata` object. In this particular example, the user would have moved the object 861 dp up (that's why it has a negative Y translation) and 180 dp to the left.

stories-interop-example.json
```diff
@@ -18,17 +29,14 @@
...
             "addedViewTextInfo": {
               "text": "this is some text",
-              "fontSizePx": 140,
-              "textColor": -1,
-              "textAlignment": 2
+              "fontSize": 14,
```
mzorz, line 32: suggest we change this from `fontSizePx` to `fontSize` where the size is expressed in `pt` (which would be absolute) or `em`? deferring to @aforcier  here, don't know what the best unit would be here

stories-interop-example.json
```diff
@@ -18,17 +29,14 @@
...
+              "textColor": "#00000000",
```
mzorz, line 33: I think this hex code format comes in handy for cross-platform text representation so, maybe a no-brainer, let me know if there's any other better options.

stories-interop-example.json
```diff
@@ -18,17 +29,14 @@
...
+              "textAlignment": "left|center|right"
```
mzorz, line 34: Probably we can do some app-level defined values like these three?

stories-interop-example.json
```diff
@@ -18,17 +29,14 @@
...
           "uri": null
         }
-      ],
-      "saveResultReason": {
-        "type": "com.wordpress.stories.compose.frame.StorySaveEvents.SaveResultReason.SaveSuccess"
-      }
+      ]
     },
     {
       "source": {
```
mzorz, line 42: Regarding this:
`- "contentUri": "content://media/external/video/media/639"`
We should probably get rid of this one altogether, it made sense for Android at the beginning since the source of a background image could be a third party app for example, that implements a Content Provider. I think given other reasons we've been discussing, we're going to have a local copy of the background image, and as such, we'll only need a `file` source here.
In the context of WordPress, such background image will also need to be uploaded to the WordPress site, and as such this will end up always being an https:// URL pointing to the background file (we could do a mediaID but the stories library should be agnostic of WordPress entities).

stories-interop-example.json
```diff
@@ -1,4 +1,15 @@
 {
+  "metadata": {
+    "revision": "1.0",
```
mzorz, line 3: this would be the revision number to correctly attempt deserialization.

stories-interop-example.json
```diff
@@ -1,4 +1,15 @@
...
+    "device": {
+      "screen": {
```
mzorz, line 5: the `device` object for now contains a screen width/height measures as informed by the device, given all other positioning measures will be relative to this canvas size.

stories-interop-example.json
mzorz: also, what looks like a good "universal" unit on iOS @bjtitus ? wonder if they save this information in Kanvas already, maybe we can borrow any ideas from there?

bjtitus: We use point sizes, typically. I believe this is the same as the "density-independent pixels" size on Android:
https://en.wikipedia.org/wiki/Point_(typography)#Apple_point

stories-interop-example.json
bjtitus: iOS has 4 built in text alignments:
- left
- right
- center
- justified
- natural

stories-interop-example.json
mzorz: ah cool! yes they even seem to have the same correlation, TIL @bjtitus  👍

stories-interop-example.json
mzorz: I see! is `natural` something to do with RTL, or is it dependent on the font?
_[searches for it]_
ah I see yes it goes with the default alignment for RTL https://developer.apple.com/documentation/uikit/nstextalignment/natural
I think if we use this we'd need to also add information about RTL to treat it the same way on every device / platform. We may want to just use maybe the first four and make sure all is contained for interoperability, or we can make this a "loose" field value and let the app figure out whatever it wants / can do with the given value here.

stories-interop-example.json
aforcier: Sounds like `sp` would be the way to go then 👍 A few more text size-related things are added with #494, most of them (shadow radius, shadow dx, shadow dy) can be in `sp` as well.

One exception would be letterspacing, which is stored in `em`s ([here's the documentation](https://developer.android.com/reference/android/widget/TextView#getLetterSpacing())), which should be an appropriate unit to keep it in.

@mzorz probably once the text feature work is done we can update this with all the new units and expand the discussion a bit - I'll follow up with you on this.

mzorz: sounds good!

PR #479, opened 2020-08-05
Normalize exported images to 9:16 - cropping resulting bitmap

### LINKED ISSUE #451: Always export 9:16 media

mzorz:
Fix #451 

Title has it, in this PR we added method `normalizeHeightExportTo916()` and crop the _resulting_ bitmap to a normalized 9:16 aspect ratio if _requested_ **and** _needed_. Given we're cropping the resulting bitmap, this means added views may be clipped as well, this risk has been properly assessed and evaluated against different approaches we could take (as for example growing the available taken space, or showing the artificially limited working area on the composer) `pbD5rk-id-p2`.

This is how it works:
1. if the screen is 16:9, we're OK and return as usual
2. if not, we resolve what the height should be for the screen's given width.
3. if the result of (2) is less than actual height, set the new target height for the cloned view to the number calculated as of (2) (cropping)
4. if the result of (2) is greater than actual height --> crop the width to match the ratio needed for the given height

**Important note**: While step/case (4) here was supposed to be less frequent to happen according to the different screen sizes we've been evaluating (*), in practice, I've found out even more common 9:16 screen sizes such as 1080x1920 still report a height value a bit smaller due to the portion of the screen the bottom Android controls take up for back arrow, home and recents navigation buttons. Hence, in order to normalize to 9:16, we need to crop on width when we find such cases.

For example, using a Pixel 3 with a screen of 1440x2792:

**Before**:
produces an image of 1440x2792 (matching the screen's resolution)
https://testmzorz3.files.wordpress.com/2020/07/wp-1595874869618.jpg

**After**:
produces an image that is now 1440x2560 (matching the target ratio)
https://testmzorz3.files.wordpress.com/2020/07/wp-1595874869642.jpg

A few more border cases:
- Pixel 3XL APi 29 is 1440x2960 - this shows the result bottom-cropped to 1440x2560 (9:16)
Screen Shot 2020-08-05 at 12 24 07

- Nexus 4 768x1280: this shows the result cropped to 666x1184, clipping some added views (this is expected)
Screen Shot 2020-08-05 at 12 27 14

- This one here 750x1334 reports an actual size of 750x1238, and as such gets cropped on width as well, resulting in 696x1238
Screen Shot 2020-08-05 at 12 29 27

- 1080x1920 reports an available height of 1794. Normalized ends up being 1009x1794:
(emulator screenshot on the right, actual produced result on the left)
Screen Shot 2020-08-05 at 12 20 49

#### To test
1. take a phone that is not 16:9
2. install the demo app and produce a slide with an image background
3. verify the output image is 16:9

#### Added notes
- (*) From https://deviceatlas.com/blog/most-used-smartphone-screen-resolutions
Screen Shot 2020-08-05 at 11 40 29

- when I mention "reported size" we're actually taking the size available to the `ghostPhotoEditorView` in full screen mode, and I was able to inspect and reproduce this is the exact same size as returned by  `[getDisplayMetrics](https://developer.android.com/reference/android/util/DisplayMetrics)` 's  `.heightPixels` and `.widthPixels`;

- Videos export will be tackled on a follow up PR

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
@@ -752,14 +752,18 @@ class PhotoEditor private constructor(builder: Builder) :
         fun onProgress(progress: Double)
     }
 
-    fun saveImageFromPhotoEditorViewAsLoopFrameFile(sequenceId: Int, photoEditorView: PhotoEditorView): File {
+    fun saveImageFromPhotoEditorViewAsLoopFrameFile(
+        sequenceId: Int,
+        photoEditorView: PhotoEditorView,
+        cropSize: Size? = null
+    ): File {
         val localFile = FileUtils.getLoopFrameFile(context, false, sequenceId.toString())
         localFile.createNewFile()
         val saveSettings = SaveSettings.Builder()
             .setClearViewsEnabled(true)
             .setTransparencyEnabled(false)
             .build()
-        FileUtils.saveViewToFile(localFile.absolutePath, saveSettings, photoEditorView)
+        FileUtils.saveViewToFile(localFile.absolutePath, saveSettings, photoEditorView, cropSize)
         return localFile
     }
 
--- a/photoeditor/src/main/java/com/automattic/photoeditor/util/BitmapUtil.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/util/BitmapUtil.kt
@@ -120,4 +120,15 @@ internal object BitmapUtil {
         v.draw(c)
         return bitmap
     }
+
+    fun createCroppedBitmapFromView(v: View, cropX: Int, cropY: Int, cropWidth: Int, cropHeight: Int): Bitmap {
+        val originalBitmap = createBitmapFromView(v)
+        return Bitmap.createBitmap(
+                originalBitmap,
+                cropX,
+                cropY,
+                cropWidth,
+                cropHeight
+        )
+    }
 }
--- a/photoeditor/src/main/java/com/automattic/photoeditor/util/FileUtils.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/util/FileUtils.kt
@@ -2,6 +2,7 @@ package com.automattic.photoeditor.util
 
 import android.content.Context
 import android.graphics.Bitmap
+import android.util.Size
 import android.view.View
 import com.automattic.photoeditor.R
 import com.automattic.photoeditor.SaveSettings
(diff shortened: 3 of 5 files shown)


aforcier:
I haven't started digging into this PR yet, but I had a thought since you also mentioned this in the width/height analytics PR:

> Important note: While step/case (4) here was supposed to be less frequent to happen according to the different screen sizes we've been evaluating (*), in practice, I've found out even more common 9:16 screen sizes such as 1080x1920 still report a height value a bit smaller due to the portion of the screen the bottom Android controls take up for back arrow, home and recents navigation buttons. Hence, in order to normalize to 9:16, we need to crop on width when we find such cases.

Have we considered hiding the navigation bar? The editing experience is meant to be immersive after all, I don't think I even realized the navigation bar was there taking up real estate until you pointed it out 😬

aforcier:
This is working great @mzorz ! :shipit:  I know you were working on possibly adjusting for the navigation bar - I'll leave it to you whether you'd prefer to add that to this PR or just merge this one and tweak in a separate PR.

Aside: I ended up fiddling around with some basic unit tests to check the logic of the `normalizeSizeExportTo916` function for myself. I think I might clean those up and add those in a PR for you to take a look at, might be nice to have the framework ready to add small test for utils like this.

mzorz:
>  I know you were working on possibly adjusting for the navigation bar - I'll leave it to you whether you'd prefer to add that to this PR or just merge this one and tweak in a separate PR.

I think I prefer to merge this one first, the navigation bar work may need some more time (and testing) 👍 

> Aside: I ended up fiddling around with some basic unit tests to check the logic of the normalizeSizeExportTo916 function for myself. I think I might clean those up and add those in a PR for you to take a look at, might be nice to have the framework ready to add small test for utils like this.

This is so cool! 🤩  - thanks for thinking about this, yes totally please I'd love to see a new PR when you have time 👍 

Thank you!

mzorz:
> Have we considered hiding the navigation bar? The editing experience is meant to be immersive after all, I don't think I even realized the navigation bar was there taking up real estate until you pointed it out

Here's a PR trying something taking into account both the collected feedback (that the navigation bar should be present to avoid confusing users) and at the same time it's semitransparent to take all of the screen (and as such, the background view will be more likely normalized to 9:16 without cropping on width): https://github.com/Automattic/stories-android/pull/481 - hope that works! 😄

PR #480, opened 2020-08-07
Elevate strings from PhotoEditor to Stories library

mzorz:
This PR copies the only 2 strings that are used for user-facing messages [from the PhotoEditor library ](https://github.com/Automattic/stories-android/blob/9c477b3e36bab8ec47abdd77182ff12045434962/photoeditor/src/main/res/values/strings.xml#L412-L413)used internally in the Stories library, so these can be extracted for translation on the host app level, or overwritten at the host app level.

See more on https://developer.android.com/studio/write/add-resources.html#resource_merging


loremattei (approved):
This looks good to me!
:shipit:

Issue #61, opened 2019-08-28
Demo 1 feedback 1.0.0.4


mzorz:
Feedback from @megsfulton on the build made out of this tag https://github.com/Automattic/portkey-android/releases/tag/v1.0.0.4-demo1

Action items written by me after a first pass on understanding what was needed and how to do things.

**Camera**
- Does the camera library we're using have the ability to tap on the view to focus on an area? I keep finding myself trying to do it and miss having that functionality

Action item:
- [x] ~add a GH ticket to do later, as a quick search tells me this is doable but may be tricky, also not sure whether this is both available on Camera2 and CameraX as well (found examples for Camera2 only).~ issue opened https://github.com/Automattic/portkey-android/issues/62

**Video**
- I like the vibrate on video capture start but it's getting picked up in the recording audio so I think unless there's a way to prevent that we should remove it. 
- For the sound toggle, can we change the label to be "Sound on" / "Sound off" ?

Action items:
- [x] ~make the video recording start when the vibration ends~ 53f5a1a
- [x] ~change sound toggle label to "Sound on" / "Sound off"~ d1022d4

**Stickers**
- I LOVE that I can swipe down to dismiss the sticker list. Can it be accessed by a swipe up from the bottom of the screen? :D
- Can the sticker list start higher up on the screen until we implement a sheet view? 
- I have a hard time making the stickers smaller than their original size, not sure what could be done to make it easier to pinch to scale them down? 

Action items:
- [x] ~implement navigation bar visible first (look further down this list), then implement swipe from bottom gesture to make the bottom sheet appear (_follow up: This is more of a nice to have, fine with making an issue for it and revisiting it later._)~ implemented in a7a0936
- [x] Option 1. make the bottom sheet slide all the way up to the top of the screen when opening or ~Option 2. maybe then to make it clearer that it's a sheet we don't extend the bg to the top. I'll mock something up for us to chat about more~.  _EDIT Made the dialog appear expanded when opening ff45ab4_
- [x] ~this one sounds tricky, investigate pinch and see if it's possible to start detecting from a bit outside the original emoji size (i.e. surrounding it), but still make it scale down~ tracked in #66 to tackle later, labeled as "enhancement"

**Text editing**
- Increase width of the done button so there's at least 20 px padding on the left and right 
- Can the initial placement of a new text object retain the same text size as in the edit mode? (And then on subsequent edits it returns to the width / height / angle the user has set.) 
[...]

PR #481, opened 2020-08-07
9:16 normalization: make image export full screen when possible

mzorz:
Following up from [this comment](https://github.com/Automattic/stories-android/pull/479#issuecomment-670215512) from @aforcier I spent a few trying to figure out if we would be able to actually get uncropped widths whenever possiblle by satisfying both these conditions:
- make the background photoview occupy the entire phone screen
- keep the navigation bar visible (so to satisfy [feedback collected here](https://github.com/Automattic/stories-android/issues/61), which is why it was being kept visible at all times), but semitransparent.

I thought if we could do this, we could take advantage of having most of the screen available.
This PR seems to work in this regard. Did the following:
```
window.setFlags(
        FLAG_TRANSLUCENT_NAVIGATION,
        FLAG_TRANSLUCENT_NAVIGATION
    )
```
[FLAG_TRANSLUCENT_NAVIGATION](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_TRANSLUCENT_NAVIGATION) is deprecated since API 30, but setting the recommended flags there will make the navigation bar hidden (which is what I was trying to avoid as per the collected feedback). Hope this won't be a problem for the time being, and I think at a later phase we can then introduce the flags to hide the navigation bar after people are already used to this screen.

With these changes, the Composer uses most of the screen, and we're able to produce 9:16 output that most of the time matches the phone screen size (or at least avoids cropping on width, as per our earlier assumptions that I wrote a caveat for in #479)

Some results:
- Pixel 3 XL - 1440x2960 (original screen size). Ends up being normalized on height to 1440x2560.
retry button needs to be adjusted as well
- 750x1334: is kept the same
- 1080x1920 is kept the same
- nexus 4 API 26: 768x1280: 720x1280

To test:
1. create story slides with the demo app
2. observe the background view occupies the full screen, and the navigation bar is shown translucent on top of the background image
3. observe the story slides selector (bottom strip) is always above the navigation bar
4. observe the add text, add emoji, delete button, etc appear above the story frame selector
5. save the slides, verify the above mentioned resolutions work as expected
6. Repeat this process with different screen sizes
7. also try the error screen by applying this gist https://gist.github.com/mzorz/852c06f673d3e27ecabab3532619c4a8
8. observe the error RETRY button  appear above the story frame selector
9. observe the story frame selector appears above the navigation bar.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -39,7 +39,8 @@
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="2147483647" />
       <option name="IMPORT_NESTED_CLASSES" value="true" />
-      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+      <option name="CONTINUATION_INDENT_IN_PARAMETER_LISTS" value="false" />
+      <option name="WRAP_EXPRESSION_BODY_FUNCTIONS" value="1" />
     </JetCodeStyleSettings>
     <XML>
       <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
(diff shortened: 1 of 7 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -39,8 +39,7 @@
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="2147483647" />
       <option name="IMPORT_NESTED_CLASSES" value="true" />
-      <option name="CONTINUATION_INDENT_IN_PARAMETER_LISTS" value="false" />
-      <option name="WRAP_EXPRESSION_BODY_FUNCTIONS" value="1" />
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
```
aforcier: If you remove `kotlin.code.style=official` from your `gradle.properties` and restore this file back to before this commit, Android Studio should stop modifying it. (Then please recommit it so there's no net change to `Project.xml`.)

I reflected this in the example file but forgot to call it out specifically: https://github.com/Automattic/stories-android/commit/9d83bc3848af337c1dc85a431abc603fe33c7af0

aforcier:
This is working great @mzorz! Glad we're able to reclaim space on 9:16 and wider devices 🎉.

This is good to go aside from the `Project.xml` changes which should be reverted. 👍 


-----

I did notice some very minor oddness on a Nexus 5X API 21 emulator. When forcing an error, the nav bar overlaps the story frame selector:

overlapping-nav-bar-nexus5x-api21

But once you hit retry and the view loads everything is fine:

overlapping-nav-bar-nexus5x-api21-2

This didn't happen on an API23 emulator or on my API29 Pixel 3. I think this is super minor and not particularly worth digging into.

.idea/codeStyles/Project.xml
mzorz: oh thanks for pointing this out - changed as indicated in d770a6a7

mzorz:
> I did notice some very minor oddness on a Nexus 5X API 21 emulator. When forcing an error, the nav bar overlaps the story frame selector:

This is interesting, I noticed that behavior consistently on other APIs and that's why the `requestLayout()` is there in https://github.com/Automattic/stories-android/pull/481/files#diff-8664737b221cd6f861bbdcd6c029b4bcR198 (and the DeleteButton as well). I think the Error Dialog is playing a role in that issue here, maybe the Dialog is displayed too soon and as such the active window is different and doesn't have the flags set, hence the inset ends up being zero. Will look into it a bit.

mzorz:
Weird, the values being set are ok, and when I set a breakpoint there I can't reproduce anymore, so it must be a timing issue. Leaving for now, confirmed couldn't reproduce on API 22, API 23, API 28 and API 29 as per my tests.

mzorz:
> This is good to go aside from the Project.xml changes which should be reverted. 

All done then! Merging 👍 :shipit:

PR #482, opened 2020-08-08
Add unit tests for 9:16 utility functions

aforcier:
While testing https://github.com/Automattic/stories-android/pull/479, I wrote a few quick unit tests to check the functionality - I'm adding them here.

I had to introduce a small data class to use instead of `android.util.Size` to make `normalizeSizeExportTo916` available for testing, since Android framework classes aren't available in the unit test environment. We could also use Robolectric and use a mocked version of the class for tests but that seemed like overkill (plus Robolectric slows down tests considerably).


aforcier:
I was going to set up our CI to run unit tests but it actually was already configured to do that - double checked that in https://github.com/Automattic/stories-android/pull/483.

mzorz (approved):
Thanks for this @aforcier ! 🙇 
:shipit:

PR #495, opened 2020-08-14
Demo app update: handle external Uris (i.e. Google Photos) - copy to local file

mzorz:
This PR updates the demo app so it can handle externally provided Uris, which are known to only be valid for a short amount of time. 

See https://github.com/wordpress-mobile/WordPress-Android/pull/12670

The difference with WPAndroid is that the demo app launches the camera capture mode right away when tapping on the + FAB, as opposed to WPAndroid where the media picker is shown first, and as such the information for Uris travels both with the new intent and is received as parameter in the creation intent in `onCreate()`, and in `onActivityResult` when the picker is shown from an ongoing Story composition.

In the Loop demo app, we only receive this information from the picker (not on creation time); the picker is launched with `startActivityForResult`, so we only need to override the `onActivityResult` method to make sure to provide the library with valid Uris, and as such the mechanisms to do so were already in place in the library / ComposeLoopFrameActivity by returning  `true`  in the `providerHandlesOnActivityResult()` override.

This PR also copies several util methods from WPAndroid  - didn't include the whole WordPressUtils library here as it would have been overkill to just utilize a few methods and have to include the library as a subtree.

To test:
1. open the demo app
2. tap on the + icon
3. capture some image (this will be the first slide)
4. now tap on the + icon on the frame selector
5. the PhotoPickerActivity shows up
6. tap on the bottom button to launch the native picker (a popup will show up: choose videos or choose pictures)
7. tap on the hamburger menu icon
8. tap on Google Photos
9. select some photos / videos depending on your choice at step 6
10. tap DONE
11. you may have to wait a bit  - then see the media items getting added as Story slides successfully.

See example video here https://cloudup.com/cgmZlSiBkQ7


aforcier:
> This PR also copies several util methods from WPAndroid - didn't include the whole WordPressUtils library here as it would have been overkill to just utilize a few methods and have to include the library as a subtree.

Haven't tried the PR yet but wanted to comment on this - we also release the utils library to bintray, we could add it as a binary instead of a subtree ([that's how WooAndroid uses it for example](https://github.com/woocommerce/woocommerce-android/blob/43ba1423f5132ba5a601e8f1ecd2d5013f70a1e2/WooCommerce/build.gradle#L144)). I think it might be worth it to save on duplication, what do you think?

mzorz:
> we also release the utils library to bintray, we could add it as a binary instead of a subtree (that's how WooAndroid uses it for example). I think it might be worth it to save on duplication, what do you think?

Good call - thanks for bringing this up 🙇  done in 24a52ac7

aforcier:
Working well, thanks Mario! :shipit:

PR #496, opened 2020-08-14
Use Uri.fromFile() insteaf of Uri.file?.absolutePath

### LINKED ISSUE #431: Crash: saving video frame
Crash spotted in alpha1 build `pr-12346-build-59863`

**Device**: Samsung s10e, Android 10

**Report:** This video was the only one captured in the Story, and it was captured using the app itself. The video was being played correctly while the user was creating the story (when adding emoji and text on top, for instance), but then crashed at the point of attempting to obtain metadata from it in the mp4composer.

Possible causes:
- Uri passed not being the same when its passed to the VideoPlayer than when it's passed down to the Mp4Composer?

Checking the code I can see we're using this for video Playing on local files (VideoPlayingBasicHandling class):
```
                val inputStream = FileInputStream(file)
                textureView.setTransform(originalMatrix)

                mediaPlayer = MediaPlayer()
                withContext(Dispatchers.IO) {
                    mediaPlayer?.setDataSource(inputStream.getFD())
                }
```

However, for mp4composer we're not doing that (check where `DataSourceUtil` is used and how).

```
--------- beginning of crash
07-08 11:56:51.616 22959 29278 E AndroidRuntime: FATAL EXCEPTION: pool-31-thread-1
07-08 11:56:51.616 22959 29278 E AndroidRuntime: Process: org.wordpress.android.prealpha, PID: 22959
07-08 11:56:51.616 22959 29278 E AndroidRuntime: java.lang.IllegalArgumentException
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at android.media.MediaMetadataRetriever.setDataSource(MediaMetadataRetriever.java:85)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at android.media.MediaMetadataRetriever.setDataSource(MediaMetadataRetriever.java:173)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at com.daasuu.mp4compose.utils.DataSourceUtil$Companion.setDataSource(DataSourceUtil.kt:25)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at com.daasuu.mp4compose.composer.Mp4Composer.getVideoResolution(Mp4Composer.kt:339)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at com.daasuu.mp4compose.composer.Mp4Composer.access$getVideoResolution(Mp4Composer.kt:24)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at com.daasuu.mp4compose.composer.Mp4Composer$start$1.run(Mp4Composer.kt:165)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
07-08 11:56:51.616 22959 29278 E AndroidRuntime: 	at java.lang.Thread.run(Thread.java:919)
```

mzorz:
Should fix #431 

While investigating this, observed the problem might have come from the content resolver not being able to figure out the right location for the file.

Given this has been reported to have crashed with a file being captured with the app itself, and we know for sure the location of the file and that it's a valid file, I figured there was a bit of information being missed when transforming the File's to `Uri` 
 by only taking its `absolutePath`.

This was indeed the case given we were doing
`Uri.parse(file.getAbsolutePath())` which is based on a simple String, while we could use `Uri.fromFile()` here as we have full control over this file. 

This is how things looked before, for a given file:

```
uri2 = {Uri$StringUri@12538} "/data/user/0/com.automattic.loop/app_tmp/tmp_wp_story1597419063035.mp4"
 authority = null
 cachedFsi = -2
 cachedSsi = -2
 fragment = null
 path = null
 query = null
 scheme = "NOT CACHED"
 ssp = null
 uriString = "/data/user/0/com.automattic.loop/app_tmp/tmp_wp_story1597419063035.mp4"
 host = "NOT CACHED"
 port = -2
 userInfo = null
 shadow$_klass_ = {Class@4668} "class android.net.Uri$StringUri"
 shadow$_monitor_ = 0
```

And this is the information we have now (observe the `scheme` in particular is correctly populated, so the content Resolver should have more information to figure out the actual content location for this Uri)

```
uri = {Uri$HierarchicalUri@12515} "file:///data/user/0/com.automattic.loop/app_tmp/tmp_wp_story1597419063035.mp4"
 authority = {Uri$Part$EmptyPart@12520} 
 fragment = {Uri$Part$EmptyPart@12521} 
 path = {Uri$PathPart@12522} 
 query = {Uri$Part$EmptyPart@12521} 
 scheme = "file"
 ssp = null
 uriString = "file:///data/user/0/com.automattic.loop/app_tmp/tmp_wp_story1597419063035.mp4"
 host = "NOT CACHED"
 port = -2
 userInfo = null
 shadow$_klass_ = {Class@2946} "class android.net.Uri$HierarchicalUri"
 shadow$_monitor_ = 0
```

Hence, we're now not losing this information and the content resolver should figure out things in a straight-forward manner.

To test:
N/A, haven't been able to reproduce myself on the devices / emulators I have, will ask someone with a Samsung S10 later.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveManager.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveManager.kt
@@ -253,7 +253,7 @@ class FrameSaveManager(
     ): Boolean {
         var callMade = false
         val uri: Uri? = (frame.source as? UriBackgroundSource)?.contentUri
-            ?: Uri.parse((frame.source as FileBackgroundSource).file?.absolutePath)
+                ?: Uri.fromFile(requireNotNull((frame.source as FileBackgroundSource).file))
         // we only need the width and height of a model canvas, not creating a canvas clone in the case of videos
         // as these are all processed in the background
         uri?.let {


stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveManager.kt
```diff
@@ -253,7 +253,7 @@ class FrameSaveManager(
     ): Boolean {
         var callMade = false
         val uri: Uri? = (frame.source as? UriBackgroundSource)?.contentUri
-            ?: Uri.parse((frame.source as FileBackgroundSource).file?.absolutePath)
+                ?: Uri.fromFile(requireNotNull((frame.source as FileBackgroundSource).file))
```
aforcier, line 256: I was wondering about using `requireNotNull` here since it means we throw an exception where we didn't before. However I see that if `uri` is `null` this function will return false and cause an exception to be thrown in `saveVideoFrame` anyhow. So I think this is fine. 👍

aforcier:
Looks good @mzorz! From a quick try I couldn't reproduce either but I didn't expect to. I did check that video saving continues to work for me, and the fix as you explained it makes sense.

If we get a new report like this we can always re-investigate.

:shipit:

PR #497, opened 2020-08-14
Added missing calls to onStoryDiscarded()

### LINKED ISSUE #421: Empty local drafts being saved
> I ended up with a bunch of local drafts, I think because tapping the back icon kept taking me out of the story editor. Maybe we need to add a confirmation when leaving the editor once you’ve started adding content?

mzorz:
Fix #421 

The actual responsibility to clear up any empty posts is already implemented on the host app side, we were just missing  a few `onStoryDiscarded()` callback calls.

To test:
CASE A: without added views
1. start a story
2. tap back or the X button
3. observe no new empty drafts are added

CASE B: with added views
1. start a story
2. add some text or emoji
3. tap back or the X button -> when the dialog appears, proceed to discard
4. observe no new empty drafts are added


aforcier:
Tested in WPAndroid, works great! :shipit:

PR #505, opened 2020-08-20
Make permission checking flag protected

### LINKED ISSUE #401: Flash of different video frame when switching slides
Steps to reproduce:

1) Add a video as the first slide (video A)
2) Add a photo slide or two
3) Add another video slide (video B)
4) Tap video A
5) Tap a photo slide
6) Tap video B - video A plays for a second before video B starts to play 

Video: https://cloudup.com/cRBqlxq7KGs

### LINKED ISSUE #501: Trying to open camera when permissions not allowed adds slides twice and doesn't trigger camera mode
When you haven't given all permissions and try capture mode for the first time when an ongoing Story is present, the already existing Slides are added twice (probably because the intent preserves the MEDIA_URIs and re-inserts them after coming back from the permissions flow).

Steps to reproduce:

First, remove all permissions. Then, follow next steps and allow each permission request as it pops up.
1. tap on the FAB and create a new story
2. pick 2 images from the picker (right the first one, don't try device picker). When prompted about image optimization, tap on LEAVE OFF.
3. now tap on + on the story composer.
4. now tap on the camera FAB, and agree o permissions
5. observe now instead of opening the camera, now you have 4 slides (because they were added twice), and the camera capture mode never appeared.

mzorz:
Fixes #501 
Related WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/12735

This PR just allows classes inheriting from ComposeLoopFrameActivity to access  `permissionsRequestForCameraInProgress` so they can hold off from calling `onLoadFromIntent()` or  act accordingly depending on whether the base class is handling permission requests.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -211,7 +211,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
     private var notificationTrackerProvider: NotificationTrackerProvider? = null
     private var prepublishingEventProvider: PrepublishingEventProvider? = null
     private var firstIntentLoaded: Boolean = false
-    private var permissionsRequestForCameraInProgress: Boolean = false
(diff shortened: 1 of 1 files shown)


aforcier:
:shipit:

PR #506, opened 2020-08-21
Using setTargetResolution instead of custom aspect ratio for VideoCapture usecase

### LINKED ISSUE #200: Video player shows stretched video
Comes from https://github.com/Automattic/portkey-android/pull/198#issuecomment-551180980

Noticed some oddness on a Pixel API26 emulator where the captured video is being stretched in edit mode (static photos are fine).

Screenshot_1573147308

Screenshot_1573147318

mzorz:
Fix #200

This seems to fix the problem with unbinding / re-binding the TextureView from the live preview use case and to the VideoCapture use case.

I have been able to reproduce the problem on the Samsung J2 and _sometimes_ on the Redmi Note 5, and with this change I could see it works 100% of the time as per my tests.
The emulators don't seem to work very well but tested on these real devices and it works alright; worth noting the Samsung J2 was having the same issue 100% of the time and now it's fixed.
- Samsung J2 OK
- Pixel 2 ANDROID 8.1 OK
- Redmi Note 5 OK
- Samsung S5e tablet: OK
- Nexus 5X: OK

To test:
1. open the demo app
2. tap on + 
3. capture a video
4. observe the live preview & the video capture keep the same resolution / aspect ratio and are not deformed
5. once the video is done recording, also observe the video player plays the video as it was recorded (no stretching)

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/camera/CameraXBasicHandling.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/camera/CameraXBasicHandling.kt
@@ -9,6 +9,7 @@ import android.os.Bundle
 import android.util.DisplayMetrics
 import android.util.Log
 import android.util.Rational
+import android.util.Size
 import android.view.ViewGroup
 import androidx.camera.core.CameraX
 import androidx.camera.core.ImageCapture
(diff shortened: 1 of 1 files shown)


aforcier:
Looks to be working well! I don't think I've ever seen the original issue so I can't confirm the fix, but tested on a Pixel 3 device and a selection of emulators and everything continues to work as normal 😀 

:shipit:

PR #507, opened 2020-08-21
Update delete slide popup menu  appearance

mzorz:
Fix #399 

**Note:** Please review and merge https://github.com/Automattic/stories-android/pull/509 first

This changes the `textAppearance` and removes the left drawable (trash icon)  for the custom "delete slide" popup menu.
I didn't add any shadows as per the referenced WPAndroid style given we are already showing a dark background to make the popupmenu stand out (and also block interaction on the rest of the screen while the menu is showing).
- [x] removes trash icon
- [x] preserves popup text style 
- [x] using a radius 1dp for corner borders to approach the style used in WPAndroid (bear in mind this here is not an actual Popup component but a custom view, was done like this because a regular popup is shown on another Window and that made the status bar appear - see #330 commit 54772da)
~- [ ] checked it works alright with dark mode~  Dark mode related work in https://github.com/Automattic/stories-android/pull/509

Screen Shot 2020-08-21 at 14 47 41

PR #509, opened 2020-08-24
Dark mode support

### LINKED ISSUE #443: Dark theme issues on WPAndroid
Tested this myself:

**Enable dark theme**
Go to Settings -> Display -> Dark theme (turn ON)

**Check the media picker when creating  a Story on WPAndroid:**
- white background on thumbnail grid.
- when in Gutenberg, add image, add media -> choose from device. The background appears WHITE as well (bug). If you choose WPMedia the background is dark. Same for Free photo library (works ok). So basically this is a bug in the Media Picker in WPAndroid. (issue on WPAndroid GH repo opened here https://github.com/wordpress-mobile/WordPress-Android/issues/12622)

**In Stories creator**
- font color white on black for inactive (no selection), blue on black when a selection is made -> same for media picker on WPAndroid normally ✅ 
- "Delete slide" option font color white over white (text can't be read)
- NEXT button blue over white, seems fine probably, unless we need to change the white for dark or another color?) cc @megsfulton

mzorz:
Fix #443 

Builds on top of #507 

#### Description
Done in this PR:

- added specific folder `values-night` overriding style for Dark Mode (read setup [here](https://material.io/develop/android/theming/dark))
- moved from `AlertDialog.Builder` to `MaterialAlertDialogBuilder`, also checked https://material.io/develop/android/components/dialogs and https://github.com/wordpress-mobile/WordPress-Android/pull/11164
- added some colors following from the WordPress Android app (may not be strictly needed given not used here, but copied since following along so the host app can override the values if needed)
- applying dark/light mode styles to `NextButton`

For further thought:
- Re: this last point, also looked into how buttons would be themed here https://github.com/material-components/material-components-android/blob/master/docs/components/Button.md, wondering whether the Next button should be a "contained button" and move away from the custom view component solution.but leaving it for a further iteration as it's not being asked for now.
- Also wondering: should we define a dark mode version / coloring for the rest of the controls? (X to close, three dots, frame selector item borders, selected indication, plus control, etc.).

#### To test
On an API 29 device / emulator with dark mode support:
- Check that things look good in both modes
- check with WPAndroid and verify it works alright
- ~[x] Known issue: when tapping on the three dots with the demo app on API 29 in dark mode only, you'll see the status bar appears. It doesn't happen on WPAndroid so, didn't look further (although it would be nice to fix of course).~ fixed in a069a880

**Light mode:**

Screen Shot 2020-08-24 at 20 48 06

Screen Shot 2020-08-24 at 20 48 08

**Dark mode:**
Screen Shot 2020-08-24 at 20 47 36

Screen Shot 2020-08-24 at 20 47 49

### DIFF
--- a/app/src/main/java/com/automattic/loop/photopicker/utils/CameraIntentUtils.java
+++ b/app/src/main/java/com/automattic/loop/photopicker/utils/CameraIntentUtils.java
@@ -90,7 +90,7 @@ public class CameraIntentUtils {
     }
 
     private static void showSDCardRequiredDialog(Context context) {
-        MaterialAlertDialogBuilder dialogBuilder = new MaterialAlertDialogBuilder(context, R.style.AlertDialogTheme);
+        MaterialAlertDialogBuilder dialogBuilder = new MaterialAlertDialogBuilder(context);
         dialogBuilder.setTitle(context.getResources().getText(R.string.sdcard_title));
         dialogBuilder.setMessage(context.getResources().getText(R.string.sdcard_message));
         dialogBuilder.setPositiveButton(context.getString(android.R.string.ok), new DialogInterface.OnClickListener() {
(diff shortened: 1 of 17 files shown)


aforcier:
Working well! Nice @mzorz :shipit:

Issue #394, opened 2020-06-26
Unexpectedly triggering delete/trash button when text is angled


megsfulton:
Steps:
1) Add text to story
2) Angle the text
3) Drag it around the screen
4) The delete button gets triggered unexpectedly 

Video: https://cloudup.com/cBtWf0bCr8W

Expected:
The delete button would only be triggered when the object is closer / visibly in contact with the delete button. 

Update: We decided to resolve this by showing the delete button on long press instead, so deleting isn't possible while simply moving the view around.

aforcier:
I had a thought about this - Canvas only triggers the trash button for individual elements when long pressing. This means that the trash action isn't available while moving a view around. Maybe we should do the same? What do you think @megsfulton ?

megsfulton:
Oh interesting. Tried it out on my iPhone and works pretty well. Let's go ahead and do that.

aforcier:
> Oh interesting. Tried it out on my iPhone and works pretty well. Let's go ahead and do that.

Great! @mzorz I'll add this to our 'hack week' group of issues then.

mzorz:
This may be improved substantially by this here https://github.com/Automattic/stories-android/pull/512 (was intending to fix some other thing but making enhancements to the detection algorithm also seems to alleviate  / address the situation described in this issue). That said, I think we should still go ahead and implement the [suggestion by Alex](https://github.com/Automattic/stories-android/issues/394#issuecomment-683188067) and totally nail the "deletion issues" for good 🤞

PR #512, opened 2020-09-02
Added views & delete view: pixel color based overlap detection

### LINKED ISSUE #339: Unable to delete large stickers
When a large sticker is scaled up it's not possible to trigger the delete button.

Video: https://cloudup.com/cmJFSuOoq-P

mzorz:
Fixes #339 

This PR implements pixel color check to see if 2 views are overlapping. Basically if both pixels located at the same place have a color other than [Color.TRANSPARENT](https://developer.android.com/reference/android/graphics/Color#TRANSPARENT), then we can be sure these two views are overlapping.

While it's not still perfect given we're calculating the dragged view bitmap with the view's matrix (but not re-creating bigger bitmaps if scaled up too much), there are some false positives with really large stickers (for example long angled text), but the experience is way better now than before and I think it's in an acceptable condition. With false positives I mean, angled text will be detected sometimes when approaching the delete view (trash icon) bounds but not visually overlapping; this is still better than not being able to delete large stickers at all as per what the issue describes. I noticed this only happens when a long text is angled to the right, but seems to work very well when angled to the left (maybe this is just a feeling).
**EDIT**: there can also be some times when detection doesn't happen right away and you have to move your finger closer to the trash icon. Not all the times and only with long boxes (long text). Again, not perfect, but way better than what we had.

[Here's a video](https://cloudup.com/cvND00yY5bN) showing this on a Samsung J2. 

To test:
1. open the demo app
2. add a view
3. drag it to the trash icon, observe it only changes to "about to delete" appearance when there's actual overlap.
4. resize, rotate, etc. and drag again, confirm you notice the improvement.

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/gesture/MultiTouchListener.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/gesture/MultiTouchListener.kt
@@ -3,7 +3,6 @@ package com.automattic.photoeditor.gesture
 import android.annotation.SuppressLint
 import android.graphics.Bitmap
 import android.graphics.Color
-import android.graphics.Point
 import android.graphics.Rect
 import android.view.GestureDetector
 import android.view.MotionEvent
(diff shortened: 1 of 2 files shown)


aforcier:
Per our discussion, putting off reviewing this one until you've had a chance to try the long press + finger-tracking approach discussed in #394, since that might supersede this PR.

aforcier:
Tested this on it own (plus merged in `develop`) and along with my attempt at fixing #394, they seem to be working well together. Will merge this and follow up with a fix to #394 later today.

:shipit:

aforcier (approved):

aforcier:
Looks like there was a crash caused by this change - I fixed it in https://github.com/Automattic/stories-android/pull/573/commits/e1e1f3de9a2c132944ee94087467c0edbf8a55c4 (part of my #573 PR to fix #394).

PR #517, opened 2020-09-14
Force non-null type in ComposeLoopFrameActivity.kt

malinajirka:
This PR wraps bundle.getString() call with `requireNotNull`.  `StorySerializerUtils.deserializeStory` accepts only non-nullable String. This change is required so we can migrate WPAndroid to SDK 29.


mzorz (approved):
Thanks for this change @malinajirka !

LGTM :shipit:

PR #519, opened 2020-09-15
Keep composed frame file & expose stories list

mzorz:
This PR adds 2 very small changes that make it easier for a host app to interact and retrieve content from serialization and the live internal StoryRepository:
- e22e671 removes the `@Transient`  annotation from `composedFrameFile` and ensures it gets serialized with `FileSerializer`, this way it's always present and can the flattened filename can be retrieved later
- c9ffa96 allows library consumers to get the full current list of Stories available in the live StoryRepository, which makes it easier for iteration when you're interested in finding all frames for any Story that match a certain criteria.


aforcier:
:shipit:

PR #520, opened 2020-09-15
Add id field to StoryFrameItem

mzorz:
This PR builds on top of #519 - together these are needed for the Story block handling on WPAndroid. 

This PR adds the `id` field (which in the context of the Stories library can be nullable and of general purpose) and exposes / adds helper methods to find and create StoryFrameItem instances easily.

To test:
N/A


aforcier:
:shipit:

PR #516, opened 2020-09-13
Fix resource references

mzorz:
Fix #515 

This PR just removes an inexistent resource reference and then adds a referenced color value to colors to the library, that may not exist in the host app.

Thanks for the heads up on this @oguzkocer ! 🙇

PR #523, opened 2020-09-15
Moved plus icon back to the stories library where it belongs

mzorz:
Fixes the plus icon being removed in #516 by mistake - and moving the resource from the example app to the stories library so it can be used in any host app

To test:
1. run the example app
2. add an image
3. observe the frame selector appears at the bottom and the adder control shows the plus icon correctly


aforcier:
:shipit:

PR #527, opened 2020-09-16
Fix story load by index

mzorz:
This PR builds on top of #520 

This PR adds a missing call to`onLoadFromInent()` in `onCreate()` which will take care of loading a pre-existing Story when it didn't come from getting a call to `onNewIntent()` (the other path would be error notifications, which worked well in that case, but no in `onCreate` since this is the first time we're going to fire up a ComposeLoopFrameActivity with an actual non empty Story to get loaded).

Also, we moved the story loading handling by index up so that case gets handled right away if the story index is being passed in the intent.

To test:
N/A, do this in conjunction with https://github.com/wordpress-mobile/WordPress-Android/pull/12939
1. open a story block that has been created elsewhere
2. tap on the story block's tool icon
3. observe the story gets loaded on the composer

Also:
- verify the other flows still work: create a new story and tap the camera icon first, create a new story and select media from device to start it with, create a new story and select media from WPMedia, etc.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -209,7 +209,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
     private var saveServiceBound: Boolean = false
     private var preHookRun: Boolean = false
     private var storyIndexToSelect = -1
-    private var storyFrameIndexToRetry: FrameIndex = StoryRepository.DEFAULT_NONE_SELECTED
+    private var storyFrameIndexToRetry: FrameIndex = StoryRepository.DEFAULT_FRAME_NONE_SELECTED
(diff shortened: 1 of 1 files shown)


stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
```diff
@@ -209,7 +209,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
     private var saveServiceBound: Boolean = false
     private var preHookRun: Boolean = false
     private var storyIndexToSelect = -1
-    private var storyFrameIndexToRetry: FrameIndex = StoryRepository.DEFAULT_NONE_SELECTED
+    private var storyFrameIndexToRetry: FrameIndex = StoryRepository.DEFAULT_FRAME_NONE_SELECTED
```
mzorz, line 212: unrelated change but spotted it and changed it

aforcier:
Set up the WPAndroid branch and tested all the flows, working well! :shipit:

PR #528, opened 2020-09-16
Added flags to allow library consumers to use temporary or permanent files for capture

mzorz:
Builds on top of #527 

This came from the need in WPAndroid to be able to reload a given unflattened Story saved elsewhere with background media that was captured using the camera capture mode of ComposeLoopFrameActivity.

Up until now, we would just use an internal temporary file to hold the media and then get rid of such file upon discarding the Story or finishing the Activity, but now we need to be able to retrieve the original media to reconstruct the Story slide properly at a later point in time.

Hence, it made sense to provide this option to the consumer of the library, which is indicated through the newly added method `setUseTempCaptureFile()` in `ComposeLoopFrameActivity`.

The flag is propagated and checked everywhere else in the library to:
- use an ordinary file (accessible from outside the app) instead of an internal directory file for capturing
- make checks everywhere where needed so to avoid cleaning up the files

If you call `setUseTempCaptureFile(false)`, no captured media will be deleted.

To test:
1. use the WPAndroid related PR for the Story block (https://github.com/wordpress-mobile/WordPress-Android/pull/12939)
2. verify you can create a story with an image captured in the stories capture mode
3. publish it
4. open the post in the editor
5. tap on the story block's tool icon
6. verify the story slide gets rendered properly (background image is not a black box, and added views show on top).

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/camera/Camera2BasicHandling.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/camera/Camera2BasicHandling.kt
@@ -720,7 +720,11 @@ class Camera2BasicHandling : VideoRecorderFragment(), View.OnClickListener {
             /**
              * create video output file
              */
-            currentFile = FileUtils.getTempCaptureFile(activity, true)
+            if (useTempCaptureFile) {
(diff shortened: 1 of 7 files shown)


aforcier:
:shipit:

PR #529, opened 2020-09-16
Show different discard dialog messaging when editing

mzorz:
This PR builds on top of https://github.com/Automattic/stories-android/pull/528

Here we added `KEY_STORY_EDIT_MODE` intent boolean extra to decide showing specific discard dialog confirmation text for a new Story or an existing one being edited.

#### Exiting the Story composer on a new story
Screen Shot 2020-09-16 at 14 16 48

#### Exiting the Story composer on a pre-existing story  being edited
Screen Shot 2020-09-16 at 14 16 16

**_Please consider the messaging and confirm this is ok / feel free to suggest a better phrasing._**

To test:
CASE A: new story
1. start a new story with a  slide
2. add a view (text, emoji)
3. tap on the cross  and observe the first dialog shows.

CASE B: existing story
1. using WPAndroid Story Block PR https://github.com/wordpress-mobile/WordPress-Android/pull/12939
2. create a story and publish it
3. go to posts list and open the newly created post
4. tap on the story block tool icon
5. observe the story loads in the composer
6. tap on the X or tap back
7. observe the second dialog shows.

PR #530, opened 2020-09-16
Introduce the GenericAnnouncementDialogProvider interface

mzorz:
This PR builds on top of #529 

This PR calls a listener when a Story has been loaded by intent using KEY_STORY_INDEX and the views are about to be presented, so this way the host app / listener can present a Dialog.

It is used in the WPAndroid app to show the "Limited editing experience" dialog depending on whether the passed Story could be completely loaded or was only re-created from the flattened images being passed as per background sources for one or more slides.

Screen Shot 2020-09-16 at 15 20 45

To test:
1. use https://github.com/wordpress-mobile/WordPress-Android/pull/12939
2. create a story and publish it
3. delete / reinstall the app (or run the app on another device) and log in
4. go to posts list and open the story post
5. tap on the story block tool icon
6. observe the story is loaded and the dialog appears as shown in the capture above.

### DIFF
--- a/app/src/main/java/com/automattic/loop/StoryComposerActivity.kt
+++ b/app/src/main/java/com/automattic/loop/StoryComposerActivity.kt
@@ -18,6 +18,8 @@ import com.automattic.loop.photopicker.RequestCodes
 import com.automattic.loop.util.CopyExternalUrisLocallyUseCase
 import com.google.android.material.snackbar.Snackbar
 import com.wordpress.stories.compose.ComposeLoopFrameActivity
+import com.wordpress.stories.compose.FrameSaveErrorDialog
+import com.wordpress.stories.compose.GenericAnnouncementDialogProvider
 import com.wordpress.stories.compose.MediaPickerProvider
 import com.wordpress.stories.compose.MetadataProvider
 import com.wordpress.stories.compose.NotificationIntentLoader
@@ -46,6 +48,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
     MetadataProvider,
     StoryDiscardListener,
     PrepublishingEventProvider,
+    GenericAnnouncementDialogProvider,
     CoroutineScope {
     private var job: Job = Job()
 
@@ -61,6 +64,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         setStoryDiscardListener(this) // optionally listen to discard events
         setPrepublishingEventProvider(this)
         setNotificationTrackerProvider(application as Loop) // optionally set Notification Tracker.
+        setGenericAnnouncementDialogProvider(this)
         // The notifiationTracker needs to be something that outlives the Activity, given the Service could be running
         // after the user has exited ComposeLoopFrameActivity
     }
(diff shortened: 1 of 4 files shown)


aforcier:
Wasn't able to test this, ran into the crash I mentioned over Slack on step 6:

```
     Caused by: kotlin.UninitializedPropertyAccessException: lateinit property EXTRA_LAUNCH_WPSTORIES_CAMERA_REQUESTED has not been initialized
        at com.wordpress.stories.compose.ComposeLoopFrameActivity$ExternalMediaPickerRequestCodesAndExtraKeys.getEXTRA_LAUNCH_WPSTORIES_CAMERA_REQUESTED(ComposeLoopFrameActivity.kt:2013)
        at com.wordpress.stories.compose.ComposeLoopFrameActivity.onLoadFromIntent(ComposeLoopFrameActivity.kt:674)
        at org.wordpress.android.ui.stories.StoryComposerActivity.onLoadFromIntent(StoryComposerActivity.kt:210)
        at com.wordpress.stories.compose.ComposeLoopFrameActivity.onCreate(ComposeLoopFrameActivity.kt:529)
        at org.wordpress.android.ui.stories.StoryComposerActivity.onCreate(StoryComposerActivity.kt:119)
        at android.app.Activity.performCreate(Activity.java:7136)
        at android.app.Activity.performCreate(Activity.java:7127)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1271)
```

mzorz:
I was able to fix the actual crash in https://github.com/wordpress-mobile/WordPress-Android/commit/8fa059d295 by moving the `setupRequestCodes` line up but this unveiled the actual problem (which if I had followed my test instructions to the letter I'd found out earlier!):
- for the case when you create a story with an old branch on the same device, the StoriesPrefs would not exist and as such it will correctly work as per the test steps (this is what I tested before).
- however if you create a story on another device (which I didn't test before and assumed it would work the same way - but it doesn't), and then try editing it with this branch on a different device, the main problem is that the `mediaIds` do not point to any actual `MediaModel` in the `MediaStore` given it hasn't been refreshed 🙃 and because of this, the Story's frames can't be re-created out of the "flattened" media.
I'll resolve this on a follow up commit on the WPAndroid PR to pre-fetch the updated media list from the site, but we'll also need to handle the case when even by prefetching we just can't find anything, and show a dialog asking for forgiveness 🙏

mzorz:
OKAY! Implemented the aforementioned paths in https://github.com/wordpress-mobile/WordPress-Android/pull/12939, see the PR description there (added two test cases C1 and C2 to cover these possibilities).

related commits:
- https://github.com/wordpress-mobile/WordPress-Android/pull/12939/commits/8fa059d29563eafbe53ea48c379fdd4232ab545b (initializing requestCodes right away to avoid the lateinit var crashes)

- UseCase refactor to prepare for testability here (not strictly related, but did it now to clean things up) https://github.com/wordpress-mobile/WordPress-Android/pull/12939/commits/df91b54a0b87a32c9c0ceb383d28b9979b9ebdef

- directly related (tackles both issues described above): https://github.com/wordpress-mobile/WordPress-Android/pull/12939/commits/448dffd9005b1be3c91d1af38e711b62f409ad25

Ready for another round @aforcier 👍

aforcier:
Working well! :shipit:

PR #542, opened 2020-09-21
Added EditMode flag passing to SaveResult

mzorz:
Title has it - part of the work needed for Stories editing processing - when a Story is edited,  a flag will be kept and shall be passed on through to the caller through the `StorySaveResult` as well.


aforcier:
:shipit:

PR #545, opened 2020-09-23
Color picker v2: 1 - Add bottom sheet

aforcier:
First (big) step of #465. Replaces the color picker RecyclerView with an (empty) bottom sheet. I'll add the actual color picker UI in a subsequent PR, but the bottom sheet behavior alone ended up being a lot of (tricky) code.

stories-bottom-sheet

I had to do some extra work to get the bottom sheet to play nice with the keyboard. Not only is there no easy way to hide the keyboard when the bottom sheet appears, but the layout recalculation when the keyboard disappears doesn't account for the bottom sheet correctly, so the text ends up being behind the bottom sheet sometimes.

The approach I went for is to manually trigger keyboard showing and hiding events when the bottom sheet is called/dismissed. Also, the view is adjusted to the bottom sheet's new height, and the height of the bottom sheet matches the keyboard's height to keep jank to a minimum.

Interested in any better ways there might be to approach all this, and more importantly if there are any breaking issues with the current behavior (at this point it's working pretty well for me).

### To test:
1. Open up the text editor, tap the color icon on the right
2. Observe that keyboard goes down, bottom sheet comes up
3. Try dismissing the bottom sheet in each of three ways:
  i. Press the color icon again
  ii. Swipe the bottom sheet down
  iii. Tap on the EditText
4. Observe that the bottom sheet goes away, and the keyboard comes back up

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/text/ColorPickerBottomSheetHandler.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/ColorPickerBottomSheetHandler.kt
@@ -60,14 +60,19 @@ class ColorPickerBottomSheetHandler(val activity: Activity, val view: View) {
         val bottomSheetCallback = object : BottomSheetBehavior.BottomSheetCallback() {
             override fun onStateChanged(bottomSheet: View, newState: Int) {
                 if (newState == BottomSheetBehavior.STATE_COLLAPSED) {
-                    // Restore the view's original size
-                    with(mainLayout.layoutParams) { height = originalViewHeight }
-
                     // Show the keyboard
                     editText.requestFocus()
                     with(activity.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager) {
                         showSoftInput(editText, InputMethodManager.SHOW_IMPLICIT)
                     }
+
+                    view.postDelayed({
+                        // Restore the view's original size, after the keyboard comes back up to avoid jerkiness
+                        with(mainLayout.layoutParams) {
+                            height = originalViewHeight
+                            mainLayout.layoutParams = this
+                        }
+                    }, BOTTOM_SHEET_DISPLAY_DELAY_MS)
                 }
             }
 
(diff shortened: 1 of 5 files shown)


mzorz:
Been testing this PR, there's something when testing on different screen sizes that will make it slightly overlap the top half of the screen, and also locking the screen then unlocking when the focus / caret was on the EditText will make the keyboard appear and push the bottom sheet all the way up, here's some videos on different devices (Samsung J2, Xaiomi Redmi Note 5, and Samsung T720 tablet)
https://cloudup.com/cl9hDfyiaaI 
Also been re-checking some recent learnings [here](https://github.com/wordpress-mobile/WordPress-Android/issues/12334#issuecomment-662669117) in the meantime to see if there's anything for this situation that is applicable as well.
Will get back at this one shortly 👍

stories/src/main/java/com/wordpress/stories/compose/text/ColorPickerBottomSheetHandler.kt
```diff
@@ -0,0 +1,142 @@
...
+                val r = Rect()
+                rootView.getWindowVisibleDisplayFrame(r)
+                val screenHeight: Int = rootView.height
+                var heightDifference = screenHeight - (r.bottom - r.top)
+
+                val resourceIdNav = activity.resources.getIdentifier("navigation_bar_height", "dimen", "android")
+                if (resourceIdNav > 0) {
+                    heightDifference -= activity.resources.getDimensionPixelSize(resourceIdNav)
+                }
+
```
mzorz, line 50: I think I found the problem of overlapping [mentioned here](https://github.com/Automattic/stories-android/pull/545#issuecomment-704267234)  is related to how the status bar is handled on different system versions. In those where the status bar doesn't "occupy" screen real estate but only show up overlapping the content behind, then the calculations seem right. But in those system versions where the status bar occupies constant space and "pushes" the screen content available to the app down, then we need to account for this amount of spaceand also offset the bottom sheet accordingly.

I tried adding the following snippet here and seems to work well:

```suggestion
                val resourceId: Int = activity.resources.getIdentifier("status_bar_height", "dimen", "android")
                if (resourceId > 0 && r.top != 0) {
                    heightDifference  -= activity.resources.getDimensionPixelSize(resourceId)
                }

```

this is without the fix:

withoutstatusbarfix

and this is with the fix:

withstatusbarfix

mzorz (dismissed):
Thanks for this work @aforcier ! It works pretty well for the most part, only found two annoying corner cases that should be handled when this [suggestion](https://github.com/Automattic/stories-android/pull/545#discussion_r500357539) is applied, and then this other[ small PR ](https://github.com/Automattic/stories-android/pull/558) gets reviewed and merged (target branch is this one, so you can see it all working together)

If you feel that other PR is good and want to merge, feel free to merge this one afterwards 👍

stories/src/main/java/com/wordpress/stories/compose/text/ColorPickerBottomSheetHandler.kt
```diff
@@ -0,0 +1,142 @@
...
+                if (heightDifference > 150) {
```
mzorz: maybe we can move this to a constant and add a comment explaining the magic number?

stories/src/main/java/com/wordpress/stories/compose/text/ColorPickerBottomSheetHandler.kt
aforcier: Thanks @mzorz! I applied your change in 216b832 (I wanted to give the variable a more explicit name).

stories/src/main/java/com/wordpress/stories/compose/text/ColorPickerBottomSheetHandler.kt
aforcier: Ah yes - done: ec3fac5

aforcier:
Thanks Mario! Addressed your suggestions and merged #558. Merging ✅

PR #546, opened 2020-09-24
Text editor: Adjust padding so style button text is always centered

### LINKED ISSUE #535: Text editing: Vertically center the title in the text styles button toggle
>Some of the font style labels sit lower in the containing rectangle than others. Strong feels like it’s not vertically centered compared to Casual, etc.

aforcier:
Fixes #535. Uses padding and one case of font size change to coax the text style button text to stay at roughly center for all fonts. (See the issue for some more detail on why simply `android:gravity="center"` doesn't cut it.)

text-style-button-center-fix

(@megsfulton has already tried a test build with these changes and given the 👍)

### To test:
Toggle the text button, make sure the text is kept at about the same height as you toggle fonts.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
@@ -17,6 +17,7 @@ import com.automattic.photoeditor.text.IdentifiableTypeface
 import com.automattic.photoeditor.text.IdentifiableTypeface.TypefaceId
 import com.wordpress.stories.R
 import java.util.TreeMap
+import kotlin.math.roundToInt
 
 /**
  * Helper class that keeps track of predefined supported text style rules and supports
@@ -126,6 +127,10 @@ class TextStyleGroupManager(val context: Context) {
         textView.setShadowLayer(textStyleRule.shadowLayer)
 
         textView.text = textStyleRule.label
+
+        // Add some corrective padding so the label for each font is kept at roughly center,
+        // despite their different intrinsic heights and placements
+        adjustTextViewLabelAlignment(typefaceId, textView)
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
```diff
@@ -135,19 +140,46 @@ class TextStyleGroupManager(val context: Context) {
         return supportedTypefaces.higherKey(typefaceId) ?: supportedTypefaces.firstKey()
     }
 
+    private fun adjustTextViewLabelAlignment(@TypefaceId typefaceId: Int, textView: TextView) {
+        // Always reset text size since it's modified for Pacifico
+        var newTextSize = 18F
```
mzorz, line 145: I was hesitant to ask to move all these values to constants but given the mission for this method is to actually update several cases to specific fixed values, having the values right there makes it much more readable. So, I think no action needed, but wanted to leave the comment here nevertheless.

stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
```diff
@@ -135,19 +140,46 @@ class TextStyleGroupManager(val context: Context) {
...
+        var paddingBottom = 0F
+        var paddingTop = 0F
+
+        when (typefaceId) {
+            TYPEFACE_ID_NUNITO -> paddingBottom = 1.9F
```
mzorz, line 150: I also thought whether there's a more generic way to center _any_ font but I guess once you know for sure which fonts you're using (which is our case) and having pre-calculated them, then assigning the given values is the best way forward (doesn't make sense to make things complicated by re-calculating them each time).
Again, no action needed, but leaving the thought to account for agreement.

mzorz (approved):
LGTM! :shipit:

stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
aforcier: Yeah I had this thought too - given how specific the task is, we're not going to reuse those constants anywhere anyway and would overall just be hurting readability. 👍

PR #547, opened 2020-09-25
Hide sticker button

aforcier:
Since we've decided the sticker feature isn't polished enough but also necessary for v1, we'll hide the button for now.

We can restore it if/when we find time to implement emoji search, extend the emoji set, and fix several 2D handling issues.


mzorz (approved):
😞 🤷‍♂️ 👍 😄 :shipit:

PR #544, opened 2020-09-22
Added FrameSaveXyz Events for broadcasting save progress with EventBus

mzorz:
This PR augments the `saveProgressListener` overrides to also send new EventBus events to signal start, progress, completion, cancelation (_currently not used_) and failure for StoryFrame save processes.

To test:
1. create a Story with a few frames
2. observe the events get sent (place breakpoints to see the code is executed)

PR #550, opened 2020-09-25
Added getCurrentStorySaveProgress() helper method

mzorz:
This PR builds on top of #544 

This adds a helper method on the StoryRepository to calculate the percentage of completion for a Frame save process, based on the existence of a `composedFrameFile` attribute on each Frame.

Also, the overall progress can be adjusted by a current in-progress addition passed by parameter, and is always calculated in proportion to the Story's frame collection size.

To Test: N/A


aforcier:
:shipit:

PR #552, opened 2020-09-30
Color picker v2: 2 - New font color selector

aforcier:
Part of the color picker v2 series. Builds on top of #545.

Moves the text color picker to the bottom sheet, and implements the new `RecyclerView` elements per the designs:

color-picker-font-color-only

Text background color will be added in the next PR.


stories/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -7,4 +7,37 @@
     android:layout_height="@dimen/color_picker_bottom_sheet_default_height"
     android:orientation="vertical">
 
+    <View
+        android:id="@+id/bottom_sheet_handle"
+        android:layout_width="40dp"
```
mzorz: can we move these to constants in dimens.xml?

stories/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -7,4 +7,37 @@
...
+        android:layout_gravity="center_horizontal"
+        android:layout_marginTop="4dp"
+        android:alpha="0.12"
+        android:background="@drawable/bg_bottom_sheet_handle" />
+
+    <TextView
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center_horizontal"
+        android:layout_marginTop="22dp"
```
mzorz: same here

stories/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -7,4 +7,37 @@
...
+        android:layout_marginBottom="4dp"
```
mzorz: also here, etc. :)

stories/src/main/res/values/arrays.xml
```diff
@@ -0,0 +1,60 @@
...
+        <item>#007017</item>
+        <item>#C08C00</item>
+        <item>#D67709</item>
+        <item>#B32D2E</item>
+        <item>#361F00</item>
+        <item>#474747</item>
+    </array>
+
+    <array name="text_colors_light">
+        <item>#D7EAFD</item>
```
mzorz, line 35: do we have code names for each of these colors? If we do, then we should probably keep the color name coding (notation) following design definition

mzorz (commented):
Left a few comments you may want to address (feel free to open an issue if time is of essence), but overall this looks good to me, it's very nice to be able to play with this feature.

One thing I noticed, but I don't think is a problem for real devices (this is only doable on the emulator AFAIU), is that if the bottom sheet is up and you still enter some text on the computer's keyboard, then the softkeyboard will be shown, pushing the bottom sheet up and covering the actual TextEdit portion of the screen where text is being entered (see gif below).

weirdnesskeyboard

If you think there may be other cases where this may happen IRL, let me know and we can figure out a workaround

aforcier:
> One thing I noticed, but I don't think is a problem for real devices (this is only doable on the emulator AFAIU), is that if the bottom sheet is up and you still enter some text on the computer's keyboard, then the softkeyboard will be shown, pushing the bottom sheet up and covering the actual TextEdit portion of the screen where text is being entered (see gif below).
> If you think there may be other cases where this may happen IRL, let me know and we can figure out a workaround

Yeah I noticed that, and I think it's just relevant to the emulator. In real life if you connect a hardware keyboard to a phone that tends to disable the software keyboard. If you somehow force both to be used at once, that's quite an edge case and it's not like the state isn't easily recoverable by dismissing the sheet or the keyboard. I think we can leave it alone.

stories/src/main/res/values/arrays.xml
aforcier: We don't really have any names (refer to pbD5rk-fT-p2). We could make up a bunch of names but that would seem excessive, if we decide we want to reuse these in some way in the future I'd say we can decide on an appropriate naming scheme/abstraction at that point.

stories/src/main/res/layout/color_picker_bottom_sheet.xml
aforcier: Moved all constants to dimens.xml in f451235 👍

aforcier:
@mzorz I merged in the base and addressed your feedback. If I missed anything let me know, I'll follow up in another PR 👍 Merging!

PR #554, opened 2020-09-30
Added onFrameRemove method to StoryDiscardListener

mzorz:
This PR adds a listener interface when the user is just about to remove a slide from the story.
Comes in handy to do any cleanup needed on the host app side.


aforcier:
:shipit:

PR #555, opened 2020-09-30
Fix composed frame file assignment

mzorz:
Title has it - we were sending a null `composedFrameFile` property with the frame when `onFrameSaveCompleted()` was being called. This PR fixes that by assigning the value before the call is made.


aforcier:
:shipit:

PR #556, opened 2020-10-02
Color picker v2: 3 - Text color background

aforcier:
Part of the color picker v2 series. Builds on top of #552.

Adds a second color picker row to the bottom sheet, and wires it up to change the text color background.

text-background-color-sheet

The initial implementation used out-of-the-box `BackgroundColorSpan`, which is pretty ugly with no padding and straight edges. Per the designs we want more space around the text and rounded corners. The first few commits though use a vanilla `BackgroundColorSpan` just to showcase the base functionality. After that, I added a custom rounded background color span (mostly based on findings from SO as noted in the file).

The text color button coloring behavior will be added in the next PR.

### A few notes:
* Per the designs, fonts that have custom shadow settings have the shadow removed when a background color is set (this ended up being needed anyway, since a text shadow with the same radius as the font padding is applied with the background color).
* There are sometimes tiny, transparent artifacts visible if you zoom way in on a multi-line text.
* Under some circumstances, the background span is cut off at the end of the text. See below:

#### Text padding cut off issue

When adding text, sometimes the end highlight in the added view gets cut off a little, so the angles are sharp instead of rounded.

Editing this text and returning, or any action that pauses the fragment (like opening the media picker and returning), seems to redraw the view and correct the problem. I've tried manually invalidating the view but that doesn't seem to do it. 

text-cutoff

As far as I can tell, this only happens if the view contains an explicit newline (a long text that gets line broken doesn't seem to count).

I think it's better to move on to other things as I'm out of ideas, but @mzorz wondering if you have any.

### To test:
* Try applying and clearing the background color (the first option in the background color recycler view)
* Check that fonts with shadows (Casual, Playful, and Bold) always have a shadow if there's no background color, and never have a shadow when there is
* Check that cycling alignment and text style works as expected


photoeditor/src/main/res/values/dimens.xml
```diff
@@ -1,5 +1,19 @@
 <resources>
-    <dimen name="text_view_padding">24dp</dimen>
+    <!-- Begin text spacing -->
+    <!-- These two sets of dimens should match, i.e. padding + horizontal_margin should equal the same for both.
+    This ensures that text is wrapped at the same point in the text editor and in the added view.
+    If using RoundedBackgroundColorSpan, changing these values may require tweaking the default padding
+    in that class so prevent any artifacts from showing up at line edges when applying the span. -->
+    <dimen name="text_view_padding">8dp</dimen>
+    <dimen name="text_view_horizontal_margin">16dp</dimen>
```
mzorz, line 8: the comment up there says these dimens should match, right? (or I'm not understanding it) - and then the actual values are not the same. Wondering if the comment needs correcting.

mzorz (approved):
Works as expected! And is really nice to play with :D 

I observed the transparent visual artifacts on the background span coloring when you zoom in, I haven't been able to make it go away when re-drawing the screen (for example by opening the media picker as suggested, see https://cloudup.com/cYBXl8Z1Ev6).

> I think it's better to move on to other things as I'm out of ideas, but @mzorz wondering if you have any.

I agree with this, this feature is so joyful to play with BTW.  Opened an issue to track it here https://github.com/Automattic/stories-android/issues/559, approving this PR - I'll let you merge it though as there may be things you want to tweak on the base PRs first. :shipit:

photoeditor/src/main/res/values/dimens.xml
aforcier: It's saying the two _sets of numbers_ should match - `text_view_*` and `edit_text_*`. So the padding + margin numbers for text view should sum to the same number as the padding + margin numbers for edit text. They do, since the edit text ones just reference the text view ones.

All this is going to change soon anyway since I'll be adding a text size slider, which will change the available screen size in the text editor, so the two numbers won't match up anymore.

aforcier:
Okay! Merged in the base, and re-used the dimens added in #552 for the new additions in 91192c8. Merging 🎉

PR #557, opened 2020-10-06
Remove text editor dialog padding on Android 11

aforcier:
After updating my phone to Android 11, I noticed the text editor screen had padding added all around it.

I haven't been able to find documentation on this anywhere - I assume it's an Android 11 thing because I've never seen it before and it still doesn't happen anywhere else.

By inspecting in Flipper, I see the padding is coming from the `DecorView` wrapping the fragment layout, so to solve it I'm setting the decor view's padding to 0 🤷‍♂️

padding-issue-beforeafter

To test:
1. On an Android 11 device, load the app without this PR and open the text editor - confirm you see the extra padding
2. With this PR, confirm padding is removed


mzorz (approved):
Nice catch there @aforcier ! 
:shipit:

PR #558, opened 2020-10-06
Hide keyboard and bottom sheet on TextEditorDialogFragment on onstop()

mzorz:
Builds on top of #545 

Prevents the issue described in https://github.com/Automattic/stories-android/pull/545#issuecomment-704267234, where leaving the bottom sheet up (but the cursor is blinking on the EditText) then locking the phone, then unlocking, would show the keyboard (because the EditText id focused) and then also push the bottom sheet up, showing both at the same time.

This PR just ensures that both are hidden when the view is about to be stopped in `onStop()`.
hidekeybwhenlock


aforcier:
Works great, thanks @mzorz !

Issue #532, opened 2020-09-16
Make it easier to increase the size of text elements with the pinch gesture


megsfulton:
In the current build, it's challenging to scale up the text using a pinch gesture. It seems like you have to be very precise. 

@alexforcier - I know we discussed possibly having the slider during text editing which will help but I think we should address the gesture issue as well.

megsfulton:
>Think I said it in a previous build. I still find selecting my text layer tricky. Feel like the tap target should be larger. Especially when there’s just one layer

^ this might also be related?

aforcier:
> I know we discussed possibly having the slider during text editing which will help but I think we should address the gesture issue as well.

Yeah it feels more necessary now - I'll try and cobble together a first pass this week.

aforcier:
The situation is improved with #561 - still keeping this open but pushing it down the priority list a bit.

PR #561, opened 2020-10-09
Text editor: Add text size slider

aforcier:
Adds a slider to the editor screen which modifies the text size.

loop-text-size-slider

This is done using a `SeekBar`. `SeekBars` aren't built to be vertical, but the common approach I found is to reuse a class from the AOSP which wraps and rotates the SeekBar, so I went with that.

(Note: There exists a newer Material component, the [Slider](https://material.io/components/sliders#theming). However it's quite new, doesn't support vertical alignment either, and seemed to misbehave when I tried it. We don't really need more than the SeekBar so I didn't go any further

This kind of addresses #532 - at least enough to make it less of an issue, and we can group it with the rest of the 2D handling issues we may address later on (or may be moot with Kanvas).

### To test:
1. Check that if you never touch the slider, default text sizes are respected for the various fonts as you cycle through them
2. Check that once the slider is used, the new size is remembered as you toggle through fonts
3. Check that the behavior of step 2 is true if you dismiss the editor and edit the resized text

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/text/TextEditorDialogFragment.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/TextEditorDialogFragment.kt
@@ -6,14 +6,11 @@ import android.graphics.Color
 import android.graphics.drawable.ColorDrawable
 import android.os.Bundle
 import android.text.Spannable
-import android.util.TypedValue
 import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
-import android.widget.SeekBar
-import android.widget.SeekBar.OnSeekBarChangeListener
 import androidx.annotation.ColorInt
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.ContextCompat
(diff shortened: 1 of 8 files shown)


stories/src/main/java/com/wordpress/stories/compose/text/TextEditorDialogFragment.kt
```diff
@@ -249,6 +261,7 @@ class TextEditorDialogFragment : DialogFragment() {
                     putInt(EXTRA_TEXT_BACKGROUND_COLOR_CODE, textStyler?.textBackgroundColor ?: Color.TRANSPARENT)
                     putInt(EXTRA_TEXT_ALIGNMENT, textStyler?.textAlignment ?: TextAlignment.default())
                     putInt(EXTRA_TYPEFACE, textStyler?.typefaceId ?: TextStyleGroupManager.TYPEFACE_ID_NUNITO)
+                    putFloat(EXTRA_TEXT_SIZE, textStyler?.fontSize ?: 0F)
```
mzorz, line 264: should we use `TextSizeSlider.Companion.TEXT_SIZE_SLIDER_MIN_VALUE` here instead of `0F`? let me know if not applicable 🙏

aforcier: Hmm good question - I don't think so. If the text styler doesn't have a font size for whatever reason, we want the text editor to use the default font size for the particular font - any fallback we set here should never actually be used. This is what is expected in [this check here](https://github.com/Automattic/stories-android/pull/561/files#diff-05151f90544e1d7f5f2e928574bcb6faR134)).

mzorz (approved):
Excellent work @aforcier 🙌 
LGTM :shipit:

PR #564, opened 2020-10-11
Added method to find a Story in the StoryRepository that contains a given frame id

mzorz:
In the host app, we may identify a Story by a specific mediaId given each flattened media produced belongs to one and only one Story. Hence, finding a Story that contains a frame with such an id should suffice to say we've found _the_ Story to which the media id belongs.

This method is needed when interacting with the host app for upload cancelation, where instead of verifying all of the mediaIds involved in a Story frame collection, we only have one mediaId being passed around (and as such should be enough to infer which Story the frame belongs to).

To test:
- N/A will provide specific use case test conditions on other PRs.


mzorz:
Going to abandon this one, realized we need after all _all_ mediaFiles Ids in order to be able to recover a Story fully from prefs because, if we only have one id and the StoryRepository doesn't contain it anymore, there's no other way we can fully recover a Story without knowing all and each of the pertaining ids. Having just one id is not enough, as we'll only be able to inflate such a frame id but won't know anything about its neighboring slides to conform the actual original Story.

Closing.

PR #570, opened 2020-10-14
Show story discard dialog even if no added views

### LINKED ISSUE #567: Xing out of a story doesn't show the dialog confirming that you'll lose your story / won't save a draft even if no elements have been placed on the screen.
What happens:
Add media to a story
Don't add text overlay
Tap X
X closes out the story editor without the confirmation dialog

What I expected:
Tapping the X always shows the confirmation dialog regardless of whether or not a user has placed an overlay on the screen

aforcier:
Fixes #567. Pressing the close button will now always show the 'Discard story post?', even if no views were added.


mzorz (approved):
:shipit: :shipit: 🚢 🇮🇹

PR #572, opened 2020-10-16
Keep added views tappable after closing discard dialog

### LINKED ISSUE #568: Bug: After delete story confirmation dialog is shown, elements on the screen are no longer tappable
What happens:
- Add media to a story
- Add a text overlay to a slide
- While on a slide with a text overlay, tap the X button
- Confirmation dialog appears
- Dismiss the confirmation dialog
- Try to move the text overlay - unable to tap on it to move or edit
- Tap on a different slide
- Tap back to the slide with the blocked text overlay - it will now be moveable

aforcier:
Fixes #568.

`addCurrentViewsToFrameAtIndex()` ends up unsetting the click listeners for added views. I don't see any reason to call that function until the user has actually confirmed the cancellation, so I moved it down. @mzorz let me know if there was a specific reason we were doing that earlier that I may have missed.

To test:
Follow the steps in #568, confirm you can now edit added views immediately after dismissing the discard dialog.


mzorz (approved):
Nice catch! 
:shipit:

PR #573, opened 2020-10-16
Long press to delete added views

### LINKED ISSUE #394: Unexpectedly triggering delete/trash button when text is angled

aforcier:
Fixes #394. You now have to long press a view for the delete button to show up - this should prevent accidental deletes while resizing or moving a view.

I started off with a flag to toggle long press state on and off, but I realized there was a simpler way by just checking if the delete view is actually visible in the collision detection logic.

**Also fixes a crash I missed before that was introduced by #512.**

### To test:
1. Confirm that you can't trigger delete by dragging added view around
2. Check that long pressing a view makes the delete button show up, and dragging to delete works as before
3. Add multiple views, delete one, and confirm that when you long press a second one the delete button shows up in the 'not ready' mode (white image, dark background)
4. Confirm that repeated deletes work
5. Check that the crash fix works: Try adding a new text view, then doing a quick tap on it. Before, this would cause a crash [here](https://github.com/Automattic/stories-android/pull/573/commits/e1e1f3de9a2c132944ee94087467c0edbf8a55c4#diff-6bc54f1069649e24610eff29aaef7b2816e4038667175b6266c615599a5eed1dL197), because the delete view hadn't been drawn yet and so the bitmap was still null.

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/gesture/MultiTouchListener.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/gesture/MultiTouchListener.kt
@@ -51,6 +51,8 @@ internal class MultiTouchListener(
     // private var onMultiTouchListener: OnMultiTouchListener? = null
     private var mOnGestureControl: OnGestureControl? = null
 
+    private var isLongPressActive: Boolean = false
+
     init {
         mScaleGestureDetector = ScaleGestureDetector(ScaleGestureListener())
         mGestureListener = GestureDetector(GestureListener())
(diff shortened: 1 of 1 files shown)


mzorz (approved):
Thank you for these changes!
LGTM :shipit:

PR #574, opened 2020-10-19
Fix discard dialog message

mzorz:
We lost the implementation of #529 at some point, bringing it back.
Also removed an unused resource while at it in 222d25c


mzorz:
Since this is bringing back functionality that was already reviewed and merged, I'm going ahead and merging - please leave a note here if you spot anything weird though @aforcier 🙇

PR #600, opened 2020-11-04
Update material library to v.1.2.1

malinajirka:
## Description
This PR updates material library version from 1.1.0 to 1.2.1. 

## How has this been tested?
I've tested several flows in WPAndroid. WPAndroid was using v 1.2.0 through a transitive dependency for some time so the udpate from 1.2.0->1.2.1 is minor ([here is a change list](https://github.com/material-components/material-components-android/releases/tag/1.2.1)).

I've tested the changes on an Emulator api 30 and api 28.


aforcier:
Tested all the flows I could think of, everything seems in order. Good to know WPAndroid's been using 1.2.0 through a transitive dependency anyway, since we've tested a lot within WPAndroid it's not too worrying.

:shipit:

aforcier (approved):

PR #601, opened 2020-11-04
Add hideCancel flag to FrameSaveErrorDialog

mzorz:
In some situations, we need an `okListener` but we don't want the `cancel` button to appear on the dialog. Added a flag to indicate so in this PR.

To test: see related WPAndroid PR with a specific use case https://github.com/wordpress-mobile/WordPress-Android/pull/13287


jd-alexander (approved):
LGTM 🚢

Issue #274, opened 2020-03-13
Text enhancements - primary thread


mzorz:
- [x] Support text alignment #186
- [x] Support changing fonts #185
- [x] Set background / foreground color for text #170
- [x] Enhanced color picker #465 
- [x] Color picker button coloring (worthy of a separate PR) --> #603 
- [x] Text size slider (improves #532)
- [x] Do a final sweep to verify dark mode support is in good shape (especially for the color picker bottom sheet)
- [x] Design review pass

![text-editing-colors-1](https://user-images.githubusercontent.com/9613966/88614197-30dd4480-d0ca-11ea-8e72-7c39f5a73684.png)

aforcier:
All done! 🎉

PR #603, opened 2020-11-07
Color picker: Update color button with text foreground/background color

aforcier:
This is the final item left for the text editor v2 feature (#274). 🎉 I had this mostly working for a while but finally took the time to clean it up and PR it.

All this does is set colorize the color picker button to match the text color, and update it every time the text color changes. The inner drop is the text color, and the surface outside the drop is the text background color.

color-picker-icon-colorized

### Implementation notes

This was surprisingly challenging because there isn't at present a way to dynamically colorize individual paths of an XML on Android. Instead you can only apply one color, which applies to all colorizable paths.

(If you take a look at `ic_textcolor.xml` you can see the paths I mean - there are 4 `path` elements, each with its own value of `android:name`.)

However, the `VectorDrawableCompat` class used to inflate XML assets actually has the capability to isolate and colorize individual paths (that's how it's able to colorize an XML with different static color values). The problem is that they haven't exposed an API to do that, and it's all handled by private functions.

I saw two options: one would be writing a custom `VectorDrawableCompat`. I gave this a try but it's a mess - it has quite a few private dependencies which we'd also have to copy, adding a ton of code that we don't really need and would have to maintain.

So I went with the second option: access the necessary `VectorDrawableCompat` functions by reflection. 🙃

This has risks of course: a future version of `VectorDrawableCompat` could remove or rename the functions, and the functionality will break. I've mitigated this as much as possible by making the reflection calls safe: there won't be a crash or anything, and instead we'll fall back to an uncolorized version of the vector asset. Since it's not a critical flow and there's an acceptable fallback, this feels like a forgivable use of reflection to me. If a future version of appcompat breaks this, we can update the reflection code when we update the library.

### To test:
1. Open up the text editor, change text foreground/background color
2. Confirm that the color picker button always reflects the colors (and the outer area is transparent when no background color is applied)
3. Confirm that when editing existing text, the color picker button is right away set to the right colors
4. You can try changing the strings of the function names we're targeting via reflection, to check that all that happens is that the icon is no longer colorized


mzorz (approved):
Really interesting! Thanks for these changes @aforcier

PR #608, opened 2020-11-13
Fix greyed out images after uninstall

### LINKED ISSUE #591: Images on the picker failing to load / freezing the app
> I’m not sure what’s up with the top three thumbnails… the ones that are solid grey.  It looks like it might be something that is failing to load?

> I tried to use one of those when making a story to see what would happen and it ended up freezing the app.

Screenshot_20201021-094127

mzorz:
Fixes #591 

This PR uses `Environment.DIRECTORY_PICTURES` to save slides so these are kept trough uninstalls

We were using [`getExternalMediaDirs`](https://developer.android.com/reference/android/content/Context#getExternalMediaDirs()) which, as the documentation states:
> This is like getExternalFilesDirs(String) in that these files will be deleted when the application is uninstalled, however there are some important differences [...]

Since Android Q we should use MediaStore API (See https://developer.android.com/training/data-storage), but given we're currently targeting API 29 we can still use this to prevent the media from being inaccessible after an uninstall.

To test:
1. using this branch: install WPAndroid, log in, create a story by capturing at least one picture with stories capture mode
2. add some text to it, save the story (publish)
3. now go to media section and tap + to add
4. observe the newly created slide can be seen there
5. uninstall the app, and reinstall
6. log in to WPAndroid
7. go to media section and tap + to add
8. observe you can now see the pictured captured with the Stories library (these are no longer greyed out meaning, these are now accessible through reinstalls).

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/util/FileUtils.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/util/FileUtils.kt
@@ -2,6 +2,7 @@ package com.automattic.photoeditor.util
 
 import android.content.Context
 import android.graphics.Bitmap
+import android.os.Environment
 import android.util.Size
 import android.view.View
 import com.automattic.photoeditor.R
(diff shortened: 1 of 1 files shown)


aforcier:
Works great! :shipit:

PR #623, opened 2020-12-04
Bring upload icon back to trigger the media picker

### LINKED ISSUE #617: Adding media to a story from the block editor opens up the camera capture view without a way to access WP or device media
Steps:

1) Add story block to post
2) Tap empty block to add media 
3) Camera capture opens without the thumbnail to access the device photos - the only option to proceed forward is to take a photo 

Is there a way to change this so that the flow is the same as creating a story from the create sheet where we show the media picker?

mzorz:
Fix #617 as per this https://github.com/Automattic/stories-android/issues/617#issuecomment-738164652

> > The thumbnail was removed long time ago but I think we can get it back for this specific case as a quick workaround for v1

> That sgtm

This PR brings the upload icon to the camera preview screen back, so there's a way to opt to open the media picker when editing an empty story block from Gutenberg.

Technically this pretty much just reverts this PR https://github.com/Automattic/stories-android/pull/447.

upload_icon

**To test**
1. use WPAndroid PR  https://github.com/wordpress-mobile/WordPress-Android/pull/13543
2. create a gutenberg blog post
3. add a Story block from the picker
4. tap on "ADD MEDIA" on the empty Story block
5. observe the camera preview mode of the Story composer appears, and the Upload thumbnail icon appears to the left of the camera shutter
6. tap on it, and the media picker should display 🎉 

Try several mixes of adding new captures, picked pictures, etc, to verify it works.


jd-alexander (commented):
Hi @mzorz thanks for these changes. I tested the behavior and the upload button was present to the left of the capture button. 

I noticed two things that I wanted to share here, to verify if they are related to this PR or not. 

1. This crash occurred when I added a video. I suspect that there might be something wrong with the emulator why it happened but I am not sure. Once the video frame was selected and the progress bar started spinning the app would crash with the error below. 

```
 java.lang.RuntimeException: setDataSource failed: status = 0xFFFFFFEA
        at android.media.MediaMetadataRetriever.setDataSource(Native Method)
        at android.media.MediaMetadataRetriever.setDataSource(MediaMetadataRetriever.java:244)
        at android.media.MediaMetadataRetriever.setDataSource(MediaMetadataRetriever.java:332)
        at org.wordpress.android.util.VideoUtils.getVideoDurationMS(VideoUtils.java:26)
        at org.wordpress.android.util.VideoUtils.getVideoDurationMS(VideoUtils.java:16)
        at org.wordpress.android.util.analytics.AnalyticsUtils.getMediaProperties(AnalyticsUtils.java:605)
        at org.wordpress.android.ui.posts.editor.EditorTracker.trackAddMediaFromDevice(EditorTracker.kt:39)
        at org.wordpress.android.ui.posts.editor.media.OptimizeMediaUseCase.optimizeMedia(OptimizeMediaUseCase.kt:61)
        at org.wordpress.android.ui.posts.editor.media.OptimizeMediaUseCase.access$optimizeMedia(OptimizeMediaUseCase.kt:21)
        at org.wordpress.android.ui.posts.editor.media.OptimizeMediaUseCase$optimizeMediaIfSupportedAsync$2$invokeSuspend$$inlined$map$lambda$1.invokeSuspend(OptimizeMediaUseCase.kt:34)
```

<kbd></kbd>

2. For the upload button I didn't see a preview in the icon. I was trying to look closely to be sure that it wasn't just a picture I took using the emulator camera that has the same shade/colors as the actual camera preview. So yes, I am just sharing this to verify that all is well as the code is probably fine and I am just experiencing some edge case. 

UPDATE : I took another look and the button is transparent so no image was set. 

<kbd></kbd>

mzorz:
Thanks for sharing your findings @jd-alexander ! I'll tackle the crash in a separate PR, however you're right it's unrelated to this PR
The transparent thumbnail is exactly like that the first time (remember this just brings funcionality we removed before but didn't change) - it's ok for now as it brings the quick access to the media picker. Next times it is run it should show the capture so, it's expected.

jd-alexander (approved):
Thanks for the comments @mzorz Since that crash will be tackled in another PR and the transparent behavior is as expected I approve this PR. LGTM 🚢 You can merge as you see fit.

.idea/compiler.xml
```diff
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
jd-alexander, line 1: Does this file need to be added to the .gitignore? Wdyt?

mzorz: it seems to be already in .gitignore as per this
```
# Intellij project files
*.iml
*.ipr
*.iws
/.idea/*

```
but this was tracked before, will check it out later separately 👍

mzorz:
Thanks for your review @jd-alexander ! 🙇 tracked the issue here https://github.com/Automattic/stories-android/issues/624

PR #625, opened 2020-12-04
Video player: catching more exceptions if we can't set the data source properly

mzorz:
When testing with different videos (especially mov and remote videos), it could be seen that trying to retrieve metadata after setting the datasource would end up resulting in various errors, that would make the app crash.

This PR aims at catching more exceptions and calling the error handler, given most of the time these errors should not crash the app and often these can be retried (for example, if the video is behind an https url, it should be possible to retry if the network conditions weren't proper at first).

To test:
1. download [this file](https://cloudup.com/cvjNmBmx7bV) to your handset / emulator (I used a Pixel 3 XL emulator with API level 29). It's 
2. try to add it in a Story from the device itself
3. observe it  doesn't crash but an error is displayed.

**BEFORE (observe it goes ANR and then crashes):**

crash_before

**AFTER (observe it does not crash any more and shows a proper error):**

crash_after


aforcier:
*cough* You can test the changes on this Pull Request by downloading the APK [here](https://5082-197228858-gh.circle-artifacts.com/0/Artifacts/Loop-pr-625-build-5082.apk).

aforcier:
Confirmed! :shipit:

PR #631, opened 2021-01-16
Added helper method to get `alt` text composed of all added text views on a given frame

mzorz:
Title has it - this PR only adds the ability to obtain a string with the text that was set for each and all added Text views for a given frame.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/story/StoryFrameItem.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/story/StoryFrameItem.kt
@@ -92,7 +92,7 @@ data class StoryFrameItem(
 
         @JvmStatic
         fun getAltTextFromFrameAddedViews(frame: StoryFrameItem): String {
-            return frame.addedViews.joinToString(separator = " ") { it -> "${it.viewInfo.addedViewTextInfo.text}" }
+            return frame.addedViews.joinToString(separator = " ") { it.viewInfo.addedViewTextInfo.text }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


stories/src/main/java/com/wordpress/stories/compose/story/StoryFrameItem.kt
```diff
@@ -89,5 +89,10 @@ data class StoryFrameItem(
                     }
             )
         }
+
+        @JvmStatic
+        fun getAltTextFromFrameAddedViews(frame: StoryFrameItem): String {
+            return frame.addedViews.joinToString(separator = " ") { it -> "${it.viewInfo.addedViewTextInfo.text}" }
```
aforcier: This can be simplified (unnecessary string template, and the arrow isn't needed either):

```suggestion
            return frame.addedViews.joinToString(separator = " ") { it.viewInfo.addedViewTextInfo.text }
```

aforcier: Since this change is pretty innocent and we'd like to get this change into the current beta, I'll just apply it and merge, forgive me Mario 🙏

aforcier:
:shipit:

stories/src/main/java/com/wordpress/stories/compose/story/StoryFrameItem.kt
mzorz: thank you! that was a nice one 👍

PR #632, opened 2021-01-19
Sending FrameSaveCompleted and FrameSaveFailed events as sticky events

mzorz:
Title has it - small change to overcome Activities / listeners getting killed on the host app.
See - https://github.com/wordpress-mobile/WordPress-Android/pull/13699

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveService.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveService.kt
@@ -329,7 +329,7 @@ class FrameSaveService : Service() {
             storySaveResult.frameSaveResult.add(FrameSaveResult(applyFrameIndexOverride(frameIndex), SaveSuccess))
 
             // dispatch FrameSaveCompleted event
-            EventBus.getDefault().post(FrameSaveCompleted(storyIndex, frameIndex, frame.id))
+            EventBus.getDefault().postSticky(FrameSaveCompleted(storyIndex, frameIndex, frame.id))
         }
 
         override fun onFrameSaveCanceled(frameIndex: FrameIndex, frame: StoryFrameItem) {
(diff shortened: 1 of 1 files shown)


jd-alexander (approved):
Thanks for these changes @mzorz I tested the WP Android behavior and the app functioned as expected with its integration with the changes made here. I also observedthe `@Subscribe` methods being marked as sticky and the EventBus that manages the `FrameSaveCompleted` and `FrameSaveFailed` events consumed them appropriately with `removeStickyEvent`. I was wondering if it would be necessary to put breakpoints in the methods for further testing in WP Android? That might not be necessary since this behavior is pretty standard with the EventBus itself, so I will approve this PR nonetheless and we can discuss that separately since that behavior is in another PR. LGTM 🚢

mzorz:
Thanks a lot for your review @jd-alexander ! 🙏 
>  I was wondering if it would be necessary to put breakpoints in the methods for further testing in WP Android? 

I did that in my case just to make sure that a sticky event didn't get processed more than once :), but totally there with you, since we're removing sticky events right at the beginning of the listeners (and no other listeners listen to these events) we're safe as we can 100% be sure the events get processed only once as intended.

mzorz:
**NOTE**: even when this PR was approved, it's better if we merged it together with the other related PRs at once, because we really want to make sure sticky events are removed in the listeners 😬 (and given I'm working on other stuff on the side, I want to avoid the risk of pointing to stories-android `develop` with this change merged, but not removing the events on the listener side)

jd-alexander:
> **NOTE**: even when this PR was approved, it's better if we merged it together with the other related PRs at once, because we really want to make sure sticky events are removed in the listeners 😬 (and given I'm working on other stuff on the side, I want to avoid the risk of pointing to stories-android `develop` with this change merged, but not removing the events on the listener side)

Yes, this is a good point. I wished there was a way for us to chain PRs and only once all the PRs have been approved would merging be possible. I'm going to ask the CEO of GitHub about that feature on Twitter 😅

PR #635, opened 2021-01-27
Remove Ruby (replacing it with the `configure` gradle plugin)

jkmassel:
This PR resolves a Ruby security warning by removing Ruby from the project entirely and replacing it with the new  `configure` gradle plugin, which we're rolling out to Android projects.

**To Test:**
- Ensure all tests pass in CI and locally – before running them locally, run `./gradlew applyConfiguration` to update the local secrets
- Download the installable build and ensure it works as expected compared to other PRs

@mzorz or @aforcier – would one of you be able to check the code changes and make sure the installable build works as expected?

@oguzkocer – could you take a pass over the gradle configuration? The fact that Android doesn't wrap its configuration in a task, but instead has to have everything present during the script evaluation pass means that we need a bunch of early returns in order for `./gradlew applyConfiguration` to have a chance to copy `gradle.properties` into position before errors are emitted. If you know of another solution I'm open to it 😅


.circleci/config.yml
```diff
@@ -54,11 +54,9 @@ jobs:
       api-version: "29"
     steps:
       - checkout
-      - bundle-install/bundle-install:
-          cache_key_prefix: installable-build
       - run:
-          name: Copy Secrets
-          command: bundle exec fastlane run configure_apply
+          name: Apply Configuration
+          command: ./gradlew applyConfiguration
```
jkmassel, line 58: No more Ruby – we can now use a `gradle` task to apply configuration secrets

.configure
```diff
@@ -1,16 +1,16 @@
 {
   "project_name": "stories-android",
   "branch": "master",
-  "pinned_hash": "746c07fd89399e5951493a32f2e853e804ca7fca",
+  "pinned_hash": "c8d5a863639968f824f2fb2a7e3e7616c828d521",
```
jkmassel, line 4: The new secrets tool also gives a security upgrade, so we need to use a different version of the repo.

Additionally, there's been a small change to `gradle.properties` in mobile secrets to reflect the new location of the release signing key.

.configure
```diff
@@ -1,16 +1,16 @@
...
   "files_to_copy": [
     {
       "file": "android/Loop/gradle.properties",
-      "destination": "gradle.properties",
+      "destination": ".configure-files/gradle.properties",
```
jkmassel: This file (and the release signing keys) are now stored in `.configure-files` rather than within the project.

The main benefit is simpler `.gitignore` rules – if all of the files live in `.configure-files/`, we can set a rule to ignore any non-encrypted file in that directory which helps prevent accidental credentials exposure.

app/build.gradle
```diff
@@ -7,6 +7,23 @@ apply plugin: 'io.sentry.android.gradle'
 android {
     compileSdkVersion rootProject.compileSdkVersion
 
+    signingConfigs {
```
jkmassel, line 10: I've explicitly set up a `signingConfig` for the installable build here – that way it can be edited from within Android Studio if needed.

app/build.gradle
```diff
@@ -7,6 +7,23 @@ apply plugin: 'io.sentry.android.gradle'
...
+        release {
+
+            if (!rootProject.gradleProperties.exists()) {
```
jkmassel: Because the `gradle.properties` file is copied into position by a gradle task, we can't guarantee that it's available at evaluation time, so we have an early exit here for that case.

app/build.gradle
```diff
@@ -84,8 +102,10 @@ dependencies {
 
 android.buildTypes.all { buildType ->
     // Add properties named "loop.xxx" to our BuildConfig
-    def inputFile = checkGradlePropertiesFile()
-    def properties = loadPropertiesFromFile(inputFile)
+    if (!rootProject.gradleProperties.exists()) {
```
jkmassel: Because the gradle.properties file is copied into position by a gradle task, we can't guarantee that it's available at evaluation time, so we have an early exit here for that case.

app/build.gradle
```diff
@@ -84,8 +102,10 @@ dependencies {
...
+        return
+    }
+    def properties = loadPropertiesFromFile(rootProject.gradleProperties)
```
jkmassel: Rather than finding this file in the `app` project, the top-level project handles finding it and providing it to any subproject if needed.

stories/build.gradle
```diff
@@ -73,8 +73,11 @@ dependencies {
 
 android.buildTypes.all { buildType ->
     // Add properties named "loop.xxx" to our BuildConfig
-    def inputFile = checkGradlePropertiesFile()
-    def properties = loadPropertiesFromFile(inputFile)
+    if (!rootProject.gradleProperties.exists()) {
```
jkmassel: Because the gradle.properties file is copied into position by a gradle task, we can't guarantee that it's available at evaluation time, so we have an early exit here for that case.

aforcier:
Confirmed the installable build is still working, and that I can build and run the project locally. Looks good to me on that score, I'll let @oguzkocer review the actual tooling changes. Thanks @jkmassel !

.configure
oguzkocer: @jkmassel My initial reaction to this was a 👍, but thinking on it more, I am not sure if I like the configuration over convention decision here. `gradle.properties` file being loaded from the root (and module) folders is the convention and Gradle does that work without any extra tasks. It's also the convention to ignore the `gradle.properties` files in git, and we already do this in this project. (some simpler ones are added for simplicity in our projects, but it's definitely not recommended) Furthermore, the `gradle.properties` file in the root will still be loaded which adds to the confusion.

I am concerned with going against with the convention and adding extra tasks + the burden of knowledge for developers that this is setup differently. If the only benefit is the simpler `.gitignore`, I don't think it's enough.

I think there is a "fancier" way to do this. We can have the encrypted properties in a `.configure-files/properties.gradle` file and load that directly in the project. However, that doesn't seem worth it to me either.

Since this affects the rest of this review, I wanted to check how strongly you were feeling about this setup and whether you'd consider reverting it to using the root `gradle.properties` file?

jkmassel: This is a great question.

Maybe there's a way to split the difference? While the `gradle.properties` file is by convention in the root of the project and added to `.gitignore`, AFAIK there's no such convention for the signing keys. I could live with keeping the `gradle.properties` file in the "correct" place if we were to avoid moving the signing keys (`.jks` file?)

WDYT?

oguzkocer: This is what I was thinking as well. I mean, I haven't really looked into the signing key part just yet and it's not something I am very experienced with, but I don't see why it shouldn't be in the `.configure-files` folder.

.circleci/config.yml
```diff
@@ -54,11 +54,9 @@ jobs:
       api-version: "29"
     steps:
       - checkout
-      - bundle-install/bundle-install:
```
oguzkocer, line 57: I think we can remove the `bundle-install` orb. We could probably remove the `Slack` and `git` orbs as well, but I am not sure why they are there in the first place and it's unrelated to this PR, so 🤷‍♂️

I gave this a try in https://github.com/Automattic/stories-android/pull/636/commits/2896c4b09f80d7d7ece4e3b2751403bd922fc365 with https://github.com/Automattic/stories-android/pull/636 and it seems to work as expected.

jkmassel: Oh nice – yeah, totally agree. Removed in 7df28b5d08b99a2070ffa4d41a583144417607b5 – we can address Slack and Git another time I think (just to avoid unrelated changes that could cause problems 😂 )

.configure
jkmassel: I've updated this in 3682896985a582c00176ed26d1d39878bdd92c85 to reflect what we discussed here.

Thanks!!

app/build.gradle
jkmassel: This has been changed back, so it doesn't work in Android Studio anymore, but it's functionally identical to what came before (though @oguzkocer – WDYT about discarding this change and the original signing stuff that was at the root of the configuration?)

oguzkocer: I don't have any experience with the signing stuff, so I am onboard with whatever you think is best 👍

oguzkocer (approved):

PR #637, opened 2021-02-02
Use task avoidance for creating build config fields

oguzkocer:
This PR contains suggestions for #635 with some minor refactors. One of the issues with creating a dependency to a file such as `gradle.properties` during the project configuration stage is that our task that wants to create this file can not be run manually. (it probably could be done automatically, but this an optional task) This issue was referenced several times by @jkmassel in these comments: [#](https://github.com/Automattic/stories-android/pull/635/files#r565840273), [#](https://github.com/Automattic/stories-android/pull/635/files#r565840333), [#](https://github.com/Automattic/stories-android/pull/635/files#r565840622)

What we need is a way to be able to run the `applyConfiguration` task without triggering the build failures that'd be caused by the missing `gradle.properties` file, but also make sure this error is handled for regular builds. And the way to achieve that is what Gradle calls [Task Configuration Avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html). Basically, we can register a task and make another task depend on it while also avoiding the evaluation of this task if it won't be run. For `applyConfiguration` task, that's exactly the case as it shouldn't depend on anything else in the build, neither should it create a dependency to the build as it's an optional task. (more on that later)

The PR also makes some minor refactors to `signingConfigs` block as we don't need a `gradleProperties` property in the root project [after moving the `gradle.properties` file to the project root](https://github.com/Automattic/stories-android/pull/635/commits/3682896985a582c00176ed26d1d39878bdd92c85).

**To test:**
1. Delete `gradle.properties` file and run `./gradlew applyConfiguration` and verify that it successfully creates the `gradle.properties` file.
2. Delete `gradle.properties` file and run `./gradlew build`, `./gradlew :stories:build` and `./gradlew :app:build` and verify that all of them tells us that we are missing the `gradle.properties` file. (Another alternative is to just run `./gradlew build --dry-run` and check that `createBuildConfigFieldsFromProperties` task would be run for both `stories` and `app` modules)
3. Delete `gradle.properties` file, run `./gradlew applyConfiguration` and `./gradlew build` and verify that everything builds successfully.

We can probably test the installable build after (if?) we merge this to the original PR.

**Further possible improvements**
1. It'd be great to use task configuration avoidance for the `configure` task so it's not evaluated unless it's directly called.
2. This might be overkill, but we can move the release signing properties to its own file as its done in the [Google docs](https://developer.android.com/studio/publish/app-signing#secure-shared-keystore). Since I made some changes there, I thought I'd bring it up, but I don't think it's worth it.
3. We can improve the error message we get from `configure` task when the `.mobile-secrets` repo is not up to date. I am currently getting the error below. It'd be good to remind developers that this might be resolved by pulling in `.mobile-secrets` repo. (This suggestion belongs to the original PR, but I wanted to group the suggestions)

```
02:06:00 [DEBUG] (1) configure::fs: Reading keys from "~/.mobile-secrets/keys.json"
thread 'main' panicked at 'Unable to decrypt and copy files: MissingProjectKey', src/configure.rs:176:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,10 +9,16 @@ android {
 
     signingConfigs {
         release {
-            storeFile file("${rootDir}/${rootProject.storeFile}")
-            storePassword rootProject.storePassword
-            keyAlias rootProject.keyAlias
-            keyPassword rootProject.keyPassword
+            if (rootProject.hasProperty("storeFile") 
+                && rootProject.hasProperty("storePassword") 
+                && rootProject.hasProperty("keyAlias") 
+                && rootProject.hasProperty("keyPassword")
+            ) { 
+                storeFile file("${rootDir}/${rootProject.storeFile}")
+                storePassword rootProject.storePassword
+                keyAlias rootProject.keyAlias
+                keyPassword rootProject.keyPassword
+            }
         }
     }
 
@@ -92,22 +98,26 @@ dependencies {
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
 }
 
-android.buildTypes.all { buildType ->
+tasks.register('createBuildConfigFieldsFromProperties') {
     // Add properties named "loop.xxx" to our BuildConfig
-    def properties = rootProject.loadGradleProperties()
-    properties.any { property ->
-    if (property.key.toLowerCase().startsWith("loop.")) {
-            buildType.buildConfigField "String", property.key.replace("loop.", "").replace(".", "_").toUpperCase(),
-                    "\"${property.value}\""
-        }
-        else if (property.key.toLowerCase().startsWith("sentry.dsn")) {
-            buildType.buildConfigField "String", property.key.replace(".", "_").toUpperCase(),
-                    "\"${property.value}\""
-        }
-        else if (property.key.toLowerCase().startsWith("loop.res.")) {
-            buildType.resValue "string", property.key.replace("loop.res.", "").replace(".", "_").toLowerCase(),
-                    "${property.value}"
+    android.buildTypes.all { buildType ->
+        def properties = rootProject.loadGradleProperties()
+        properties.any { property ->
+        if (property.key.toLowerCase().startsWith("loop.")) {
+                buildType.buildConfigField "String", property.key.replace("loop.", "").replace(".", "_").toUpperCase(),
+                        "\"${property.value}\""
+            }
+            else if (property.key.toLowerCase().startsWith("sentry.dsn")) {
+                buildType.buildConfigField "String", property.key.replace(".", "_").toUpperCase(),
+                        "\"${property.value}\""
+            }
+            else if (property.key.toLowerCase().startsWith("loop.res.")) {
+                buildType.resValue "string", property.key.replace("loop.res.", "").replace(".", "_").toLowerCase(),
+                        "${property.value}"
+            }
         }
     }
 }
+
+preBuild.dependsOn(tasks.named("createBuildConfigFieldsFromProperties"))
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,18 +9,10 @@ android {
 
     signingConfigs {
         release {
-
-            if (!rootProject.gradleProperties.exists()) {
-                println "Unable to read ${gradleProperties}"
-                return
-            }
-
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -9,18 +9,16 @@ android {
 
     signingConfigs {
         release {
-
-            if (!rootProject.gradleProperties.exists()) {
-                println "Unable to read ${gradleProperties}"
-                return
+            if (rootProject.hasProperty("storeFile")
```
jkmassel, line 12: I'm not crazy about this (I like your suggestion of moving it to a separate file!) but I don't feel _strongly_ about it either 🤷‍♂️

IMHO the check for whether these exist is helpful in release mode where we should error our if it's missing, rather than skipping all this (though I'd assume it'll complain that the signing configuration is set up incorrectly, so it may not matter).

Not a big deal for this – your way is more aligned with the original project so I think it's how we should go for now, but definitely room for improvement! :)

app/build.gradle
```diff
@@ -100,32 +98,26 @@ dependencies {
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
 }
 
-android.buildTypes.all { buildType ->
+tasks.register('createBuildConfigFieldsFromProperties') {
```
jkmassel, line 101: This is a _much_ better way to do this – I wasn't sure how to wrap it in a task, but I'm super happy to see you were able to!

jkmassel (approved):
I really, really like this approach – it's far better than what I'd come up with!

Regarding your feedback:

> It'd be great to use task configuration avoidance for the configure task so it's not evaluated unless it's directly called.

I'd tend to agree, though the benefit here is you get the new secrets applied automatically after switching branches (which is usually what you'd want I think?). I could go either way on it tbh.

> This might be overkill, but we can move the release signing properties to its own file as its done in the Google docs. Since I made some changes there, I thought I'd bring it up, but I don't think it's worth it.

I like this idea but agree it's not worth it in the context of _this_ PR (especially because this app doesn't go to the store on its own, but rather as part of a larger one).

> We can improve the error message we get from configure task when the .mobile-secrets repo is not up to date. I am currently getting the error below. It'd be good to remind developers that this might be resolved by pulling in .mobile-secrets repo. (This suggestion belongs to the original PR, but I wanted to group the suggestions)

Yes, I would tend to agree with that – we could definitely improve it [upstream](https://github.com/automattic/configure) and benefit all the apps using it.

app/build.gradle
oguzkocer: > I'm not crazy about this (I like your suggestion of moving it to a separate file!) but I don't feel strongly about it either 🤷‍♂️

I am not a big fan of it either, but I didn't have a solution in mind that I really liked.

> IMHO the check for whether these exist is helpful in release mode where we should error our if it's missing, rather than skipping all this (though I'd assume it'll complain that the signing configuration is set up incorrectly, so it may not matter).

As you said, it should give an error when we are trying the sign, so.. I thought this was the harmless way to do it.

> Not a big deal for this – your way is more aligned with the original project so I think it's how we should go for now, but definitely room for improvement! :)

👍

oguzkocer:
> I'd tend to agree, though the benefit here is you get the new secrets applied automatically after switching branches (which is usually what you'd want I think?). I could go either way on it tbh.

I don't think that's the case here. We _could_ make that happen, but it would not work for developers without the `mobile-secrets` repo and I am not sure if it'd be intuitive. One tricky bit with `gradle.properties` is that, you can add local only properties to it, such as the desired memory allocation. So, I personally wouldn't like my file to be changed without me being aware of it (I also don't like auto-magical stuff very much), but I might be the odd one out. 🤷‍♂️ 

If we don't want to implement that ^, I think it makes sense to have task configuration avoidance for the `applyConfiguration` task. Benefit to that would be fairly small for one app, but considering we'll have this everywhere, it might be worth it.

> I like this idea but agree it's not worth it in the context of this PR (especially because this app doesn't go to the store on its own, but rather as part of a larger one).

👍 

> Yes, I would tend to agree with that – we could definitely improve it upstream and benefit all the apps using it.

That'd be a nice quality of life improvement 👍

PR #639, opened 2021-02-12
Move font loading to background thread

### LINKED ISSUE #13693: ApplicationNotResponding: in FontResolver in Stories
Sentry Issue: [WORDPRESS-ANDROID-120C](https://sentry.io/share/issue/796af984441d4fb3b4fdefbcdda55dbb/)

This is happening when loading fonts in Stories, perhaps we can try force pre-loading them somehow in the background instead of waiting until the TextDialog appears (when the user needs it)?

```
io.sentry.android.core.ApplicationNotResponding: Application Not Responding for at least 5000 ms.
    at android.os.BinderProxy.transactNative(BinderProxy.java)
    at android.os.BinderProxy.transact(BinderProxy.java:578)
    at android.content.ContentProviderProxy.query(ContentProviderNative.java:421)
    at android.content.ContentResolver.query(ContentResolver.java:951)
    at android.content.ContentResolver.query(ContentResolver.java:887)
    at androidx.core.provider.FontsContractCompat.getFontFromProvider(FontsContractCompat.java:835)
    at androidx.core.provider.FontsContractCompat.fetchFonts(FontsContractCompat.java:733)
    at androidx.core.provider.FontsContractCompat.getFontInternal(FontsContractCompat.java:184)
    at androidx.core.provider.FontsContractCompat.getFontSync(FontsContractCompat.java:243)
    at androidx.core.graphics.TypefaceCompat.createFromResourcesFamilyXml(TypefaceCompat.java:117)
    at androidx.core.content.res.ResourcesCompat.loadFont(ResourcesCompat.java:426)
    at androidx.core.content.res.ResourcesCompat.loadFont(ResourcesCompat.java:372)
    at androidx.core.content.res.ResourcesCompat.getFont(ResourcesCompat.java:241)
    at com.wordpress.stories.compose.text.TextStyleGroupManager$Companion.getFontWithFallback(TextStyleGroupManager.kt:236)
    at com.wordpress.stories.compose.text.TextStyleGroupManager$Companion.resolveTypeface(TextStyleGroupManager.kt:229)
    at com.wordpress.stories.compose.text.TextStyleGroupManager$Companion.access$resolveTypeface(TextStyleGroupManager.kt:210)
    at com.wordpress.stories.compose.text.TextStyleGroupManager.getFont(TextStyleGroupManager.kt:108)
    at com.wordpress.stories.compose.text.TextStyleGroupManager.<init>(TextStyleGroupManager.kt:99)
    at com.wordpress.stories.compose.text.TextEditorDialogFragment.onAttach(TextEditorDialogFragment.kt:54)
    at androidx.fragment.app.Fragment.performAttach(Fragment.java:2673)
    at androidx.fragment.app.FragmentStateManager.attach(FragmentStateManager.java:263)
    at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1182)
    at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1368)
    at androidx.fragment.app.FragmentManager.moveFragmentToExpectedState(FragmentManager.java:1446)
    at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1509)
    at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:447)
    at androidx.fragment.app.FragmentManager.executeOps(FragmentManager.java:2181)
    at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:2004)
    at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1959)
    at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1861)
    at androidx.fragment.app.FragmentManager$4.run(FragmentManager.java:413)
    at android.os.Looper.loop(Looper.java:237)
    at android.app.ActivityThread.main(ActivityThread.java:7948)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1075)```

### PR DESCRIPTION
**Pull request by aforcier**
Should fix https://github.com/wordpress-mobile/WordPress-Android/issues/13693.

Stories relies on several downloadable fonts. It seems the problem was that on slow connections (or perhaps if the Google font server was unavailable to the user for any reason), obtaining the fonts can take a while, and we were making the request on the UI thread - leading to an ANR.

`ResourcesCompat#getFont()` turns out to support asynchronous font requests for this reason. However, asynchronously downloading when the text editor launches is too late; we'd always be using the fallback fonts instead of the custom fonts on first run of the story feature, which we don't want.

My solution is to cue the font loading ahead of time, when the story creator is started up. This will trigger the font download (if they're not already downloaded - otherwise `ResourcesCompat` will serve up the cached local font reference) and save the resolved font reference in the new `FontRequester`. Later requests to display the font are handled by `FontRequester`, which will serve up whatever it got from `ResourcesCompat` (or the fallback font in case of errors or if the callback hasn't been called yet).

Note: `FontRequester` behaves like a singleton - it should really be a dagger module and injected into `ComposeLoopFrameActivity` and `TextStyleGroupManager`, but that was starting to become too big a diversion. I think this is okay for now and we should plan a bigger dagger refactor for this and other areas where we need it (see #357 also).

### To test

Try things out on an emulator wiped clean (so any locally saved fonts are removed). If you filter for `font` in the log and run the app, you should see this when the creator is launched:

(The `V` logs from `com.automattic.loop` are temporary logging I added for testing and not included in the PR - the first set inside `registerFont()`, the second set in `onFontRetrieved()`).

```
com.automattic.loop V/FontRequester: Registering font: 2131296257
com.automattic.loop V/FontRequester: Registering font: 2131296256
com.automattic.loop V/FontRequester: Registering font: 2131296258
com.automattic.loop V/FontRequester: Registering font: 2131296259
com.automattic.loop V/FontRequester: Registering font: 2131296261
com.automattic.loop V/FontRequester: Registering font: 2131296260
com.google.android.gms I/FontsContentProvider: Received query name=Nunito&weight=700, URI content://com.google.android.gms.fonts
com.google.android.gms I/FontsContentProvider: Query [name=Nunito&weight=700] resolved to {Nunito, wdth 100.0, wght 700, ital 0.0, bestEffort false}
com.google.android.gms I/FontsContentProvider: Fetch {Nunito, wdth 100.0, wght 700, ital 0.0, bestEffort false} end status Status{statusCode=SUCCESS, resolution=null}
com.google.android.gms I/FontsContentProvider: Pulling font file for id = 6, cache size = 5
com.google.android.gms I/FontsContentProvider: Received query Libre Baskerville, URI content://com.google.android.gms.fonts
com.google.android.gms I/FontsContentProvider: Query [Libre Baskerville] resolved to {Libre Baskerville, wdth 100.0, wght 400, ital 0.0, bestEffort false}
com.google.android.gms I/FontsContentProvider: Fetch {Libre Baskerville, wdth 100.0, wght 400, ital 0.0, bestEffort false} end status Status{statusCode=SUCCESS, resolution=null}
com.google.android.gms I/FontsContentProvider: Pulling font file for id = 7, cache size = 6
com.google.android.gms I/FontsContentProvider: Received query Pacifico, URI content://com.google.android.gms.fonts
com.google.android.gms I/FontsContentProvider: Query [Pacifico] resolved to {Pacifico, wdth 100.0, wght 400, ital 0.0, bestEffort false}
com.google.android.gms I/FontsContentProvider: Fetch {Pacifico, wdth 100.0, wght 400, ital 0.0, bestEffort false} end status Status{statusCode=SUCCESS, resolution=null}
com.google.android.gms I/FontsContentProvider: Pulling font file for id = 8, cache size = 7
com.google.android.gms I/FontsContentProvider: Received query name=Space Mono&weight=700, URI content://com.google.android.gms.fonts
com.google.android.gms I/FontsContentProvider: Query [name=Space Mono&weight=700] resolved to {Space Mono, wdth 100.0, wght 700, ital 0.0, bestEffort false}
com.google.android.gms I/FontsContentProvider: Fetch {Space Mono, wdth 100.0, wght 700, ital 0.0, bestEffort false} end status Status{statusCode=SUCCESS, resolution=null}
com.google.android.gms I/FontsContentProvider: Pulling font file for id = 9, cache size = 6
com.google.android.gms I/FontsContentProvider: Received query Shrikhand, URI content://com.google.android.gms.fonts
com.google.android.gms I/FontsContentProvider: Query [Shrikhand] resolved to {Shrikhand, wdth 100.0, wght 400, ital 0.0, bestEffort false}
com.google.android.gms I/FontsContentProvider: Fetch {Shrikhand, wdth 100.0, wght 400, ital 0.0, bestEffort false} end status Status{statusCode=SUCCESS, resolution=null}
com.google.android.gms I/FontsContentProvider: Pulling font file for id = 10, cache size = 7
com.automattic.loop V/FontRequester: Font retrieved from ResourcesCompat: 2131296258
com.automattic.loop V/FontRequester: Font retrieved from ResourcesCompat: 2131296257
com.automattic.loop V/FontRequester: Font retrieved from ResourcesCompat: 2131296256
com.automattic.loop V/FontRequester: Font retrieved from ResourcesCompat: 2131296259
com.automattic.loop V/FontRequester: Font retrieved from ResourcesCompat: 2131296261
com.automattic.loop V/FontRequester: Font retrieved from ResourcesCompat: 2131296260
```

If after that you open the text editor, you should see the fonts already preloaded, and no new requests to `FontsContentProvider` in the log.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
@@ -8,6 +8,7 @@ import androidx.annotation.FontRes
 import androidx.core.content.res.ResourcesCompat
 import androidx.core.content.res.ResourcesCompat.FontCallback
 import androidx.core.provider.FontsContractCompat.FontRequestCallback.FontRequestFailReason
+import java.util.Collections
 
 /**
  * Handles dispatching requests to ResourcesCompat.getFont() asynchronously. These requests can take a long time,
@@ -22,7 +23,7 @@ import androidx.core.provider.FontsContractCompat.FontRequestCallback.FontReques
 object FontRequester {
     private val TAG = FontRequester::class.java.simpleName
 
-    val fontMap = mutableMapOf<Int, Typeface>()
+    val fontMap: MutableMap<Int, Typeface> = Collections.synchronizedMap(mutableMapOf<Int, Typeface>())
 
     fun registerFont(context: Context, @FontRes fontRes: Int, fallback: Typeface) {
         // Set the font to the fallback right away in case it's requested before getFont completes.
--- a/stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
@@ -1,0 +1,72 @@
+package com.wordpress.stories.compose.text
+
+import android.content.Context
+import android.content.res.Resources.NotFoundException
+import android.graphics.Typeface
+import android.util.Log
+import androidx.annotation.FontRes
+import androidx.core.content.res.ResourcesCompat
+import androidx.core.content.res.ResourcesCompat.FontCallback
+import androidx.core.provider.FontsContractCompat.FontRequestCallback.FontRequestFailReason
+
+/**
+ * Handles dispatching requests to ResourcesCompat.getFont() asynchronously. These requests can take a long time,
+ * for example when the font is downloadable and there are connection issues.
+ *
+ * When registering or requesting a font from FontRequester, a fallback Typeface must be specified which will
+ * be returned if:
+ * - The request to ResourcesCompat hasn't completed yet
+ * - The request to ResourcesCompat failed for any reason
+ * - The font being requested wasn't previously registered with FontRequester
+ */
+object FontRequester {
+    private val TAG = FontRequester::class.java.simpleName
+
+    val fontMap = mutableMapOf<Int, Typeface>()
+
+    fun registerFont(context: Context, @FontRes fontRes: Int, fallback: Typeface) {
+        // Set the font to the fallback right away in case it's requested before getFont completes.
+        fontMap[fontRes] = fallback
+        try {
+            ResourcesCompat.getFont(context, fontRes, fontCallbackFor(context, fontRes), null)
+        } catch (e: NotFoundException) {
+            Log.e(TAG, "Font ${context.resources.getResourceEntryName(fontRes)} not found")
+        }
+    }
+
+    fun getFont(@FontRes fontRes: Int, fallback: Typeface): Typeface {
+        return fontMap[fontRes] ?: fallback
+    }
+
+    private fun fontCallbackFor(context: Context, @FontRes fontRes: Int): FontCallback {
+        return object : FontCallback() {
+            /**
+             * Called when an asynchronous font was finished loading.
+             *
+             * @param typeface The font that was loaded.
+             */
+            override fun onFontRetrieved(typeface: Typeface) {
+                fontMap[fontRes] = typeface
+            }
+
+            /**
+             * Called when an asynchronous font failed to load.
+             *
+             * @param reason The reason the font failed to load. One of
+             * [FontRequestFailReason.FAIL_REASON_PROVIDER_NOT_FOUND],
+             * [FontRequestFailReason.FAIL_REASON_WRONG_CERTIFICATES],
+             * [FontRequestFailReason.FAIL_REASON_FONT_LOAD_ERROR],
+             * [FontRequestFailReason.FAIL_REASON_SECURITY_VIOLATION],
+             * [FontRequestFailReason.FAIL_REASON_FONT_NOT_FOUND],
+             * [FontRequestFailReason.FAIL_REASON_FONT_UNAVAILABLE] or
+             * [FontRequestFailReason.FAIL_REASON_MALFORMED_QUERY].
+             */
+            @Suppress("KDocUnresolvedReference")
+            override fun onFontRetrievalFailed(@FontRequestFailReason reason: Int) {
+                // Just log the failure - the font mapping is already pointed to the fallback font.
+                Log.e(TAG, "Font retrieval for ${context.resources.getResourceEntryName(fontRes)} failed. " +
+                        "FontRequestFailReason error code: $reason.")
+            }
+        }
+    }
+}
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -341,6 +341,9 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
             }
         }
 
+        // Pre-load the custom fonts if necessary
+        TextStyleGroupManager.preloadFonts(this)
+
         workingAreaRect = calculateWorkingArea()
         photoEditor = PhotoEditor.Builder(this, photoEditorView)
             .setPinchTextScalable(true) // set flag to make text scalable when pinch
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -430,7 +430,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
 
         photoEditor.setFontResolver(object : FontResolver {
             override fun resolve(@TypefaceId typefaceId: Int): IdentifiableTypeface {
-                return TextStyleGroupManager.getIdentifiableTypefaceForId(typefaceId, this@ComposeLoopFrameActivity)
+                return TextStyleGroupManager.getIdentifiableTypefaceForId(typefaceId)
             }
         })
 
--- a/stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/TextStyleGroupManager.kt
@@ -1,7 +1,6 @@
 package com.wordpress.stories.compose.text
 
 import android.content.Context
-import android.content.res.Resources.NotFoundException
 import android.graphics.Typeface
 import android.util.TypedValue
 import android.widget.TextView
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt [hunk 1]
```diff
@@ -0,0 +1,72 @@
...
+ * - The font being requested wasn't previously registered with FontRequester
+ */
+object FontRequester {
+    private val TAG = FontRequester::class.java.simpleName
+
+    val fontMap = mutableMapOf<Int, Typeface>()
+
+    fun registerFont(context: Context, @FontRes fontRes: Int, fallback: Typeface) {
+        // Set the font to the fallback right away in case it's requested before getFont completes.
+        fontMap[fontRes] = fallback
```
mzorz, line 30: nice touch here acting preemptively ✅

stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
```diff
@@ -0,0 +1,72 @@
...
+        try {
+            ResourcesCompat.getFont(context, fontRes, fontCallbackFor(context, fontRes), null)
+        } catch (e: NotFoundException) {
+            Log.e(TAG, "Font ${context.resources.getResourceEntryName(fontRes)} not found")
```
mzorz, line 34: This is a nitpick that ended up being an interesting internal check  - giving a heads up so you can decide whether this makes sense to consider or not 👍:

I was going to ask whether it is possible to get an Exception from an async call directly 😅  but then realized looking into ResourcesCompat's  source code that `getFont` calls `loadFont()` internally and it only throws this when no callback is passed
```
        if (typeface == null && fontCallback == null) {
            throw new NotFoundException("Font resource ID #0x"
                    + Integer.toHexString(id) + " could not be retrieved.");
        }
```
Even `getFont()` has a [precheck](https://android.googlesource.com/platform/frameworks/support/+/HEAD/core/core/src/main/java/androidx/core/content/res/ResourcesCompat.java#358) that the callback shouldn't be null or it throws an NPE:

```
Preconditions.checkNotNull(fontCallback);
```

`Preconditions.checkNotNull` will throw a NPE if the passed parameter is null - I wondered if we should also catch  NPE here in this try/catch clause, but I guess in our case it will never be thrown because we always pass a callback which is also an `object` method so it is guaranteed to live with the Application process.

So yes, all this to say, we can probably get rid of the try/catch clause and only declare the `throws ...` at the method declaration level and let the user of the library decide. Unless we want to encapsulate this behavior and just have the FontRequester resolve everything, which is also nice.

stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
```diff
@@ -0,0 +1,72 @@
...
+ * When registering or requesting a font from FontRequester, a fallback Typeface must be specified which will
+ * be returned if:
+ * - The request to ResourcesCompat hasn't completed yet
+ * - The request to ResourcesCompat failed for any reason
+ * - The font being requested wasn't previously registered with FontRequester
+ */
+object FontRequester {
+    private val TAG = FontRequester::class.java.simpleName
+
+    val fontMap = mutableMapOf<Int, Typeface>()
```
mzorz: I've seen Kotlin provides a [`LinkedHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html) as the implementation of `mutableMapOf`.
A LinkedHashMap is not thread safe, so I wondered what would happen in the (possibly very rare) case where several fonts are retrieved and then their async callbacks decide to assign values to it (write) at the same time.

Generally I'd think there wouldn't be a problem given all keys in the map are guaranteed to be different, but it really depends on the implementation (maybe registering a new key makes values to shift in the underlying data structure, which would then potentially be problematic given the insertion method is not synchronized).

So in order to avoid this, this would need to be externally synchronized, one way is to use `Collections.synchronizedMap`  wdyt?

```suggestion
    val fontMap = Collections.synchronizedMap(mutableMapOf<Int, Typeface>())
```

mzorz (changes requested):
Nice work @aforcier ! 🙌 

I tested and verified it works, on API 22 and API29, with airplane mode ON first and OFF afterwards to verify it loads.

I love how code looks simple to understand and is well documented - left a couple of comments, one of them I think would be nice to address before merging (re: the thread safety of the `fontMap`). 

Feel free to merge afterwards 👍

stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
aforcier: > Preconditions.checkNotNull will throw a NPE if the passed parameter is null - I wondered if we should also catch NPE here in this try/catch clause, but I guess in our case it will never be thrown because we always pass a callback which is also an object method so it is guaranteed to live with the Application process.

Yes agreed. I think it makes sense to continue to catch `NotFoundException` since it's explicitly thrown by `getFont` (and even though looking into the internals reveals that it shouldn't get thrown in our case, the internals could always change in a future update).

> Unless we want to encapsulate this behavior and just have the FontRequester resolve everything, which is also nice.

I think we can encapsulate the behavior - in general `FontRequester` is supposed to give you back the fallback font if anything goes wrong, I think eating the potential `NotFoundException` and using the fallback falls within what you'd expect the object to do.

stories/src/main/java/com/wordpress/stories/compose/text/FontRequester.kt
aforcier: Good point @mzorz , let's be safe 👍 Made the change in 5e79486 (the `fontMap` type needs an explicit cast for null safety reasons, otherwise it generates a warning).

aforcier:
Thanks @mzorz! Updated and merged per your comment. WPAndroid PR updating the hash is here: https://github.com/wordpress-mobile/WordPress-Android/pull/14065

PR #647, opened 2021-02-24
Pan / zoom background images

### LINKED ISSUE #646: Pinch to zoom on background images / videos
The flow should work like this:
1. add images, videos, slides, etc
2. once you land on the composer, by default pinch to zoom will work on the background image ~or video~
~3. as soon as you add views, pinch will only work on the added views~
~4. as soon as the added views are gone if you delete all of them, pinch to zoom on the background will work again.~

mzorz:
Fix #646 for `IMAGE` background frames (not videos)

This PR adds pinch to zoom for background images

**Note**: we didn't disable the pinch gesture detection as specified in the issue, because it felt it didn't bother the intent of dragging / pinching on added views at all for me. Let me know if we still want to be strict about that and I can follow the described behavior enabling it when no added views are present, and disabling it if any added views exist on the frame.

https://user-images.githubusercontent.com/6597771/109062297-83192f00-76c6-11eb-8441-9710f36ef366.mp4

#### Description
The [PhotoView](https://github.com/Baseflow/PhotoView) library which is already being used in [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/search?q=photoview) provides seamless pinch to zoom handling. What we needed to do is to provide a way to "remember" the underlying ImageView's [`imageMatrix`](https://developer.android.com/reference/android/widget/ImageView#getImageMatrix()) in between Story frame switches and re-applying the defined matrix on the `ghostPhotoEditor`  when painting each frame in FrameSaveManager.

This is how it works:
- Each time `onStoryFrameSelected()` is called, we save the current matrix state to the frame's `BackgroundSource`'s `backgroundViewInfo` object which holds an flattened array of such 3x3 matrix, so we can reload it later when we need to show it again.
- Then, we perform business as usual and load the newly selected frame on the PhotoEditor, with the caveat that we need to wait for the image drawable to be actually set on the underlying ImageView before we can apply its "remembered" matrix. For this, we use Glide's [`RequestListener.onResourceReady()`](https://bumptech.github.io/glide/javadocs/440/com/bumptech/glide/request/RequestListener.html#onResourceReady) interface, and at that point we issue a task on the main thread using a Handler. We need this because `onResourceReady` callback method is called right before the drawable is set (in fact, we must return `false` so Glide does actually end up setting it after we exit the callback). Now in the next UI cycle we can be sure the [drawable is set](https://github.com/Baseflow/PhotoView/blob/139a9ffeaf70bd628b015374cb6530fcf7d0bcb7/photoview/src/main/java/com/github/chrisbanes/photoview/PhotoViewAttacher.java#L279-L289), and as such we can go ahead an reapply the `imageMatrix` there.
- When you tap on the check icon on the top-right of the screen to save your frames, we use the information held by `backgroundViewInfo` on each frame's `BackgroundSource` to rebuild and apply the matrix on the `ghostPhotoEditorView` we use to paint and save each bitmap to a picture file.

#### Note on serialization
~I haven't tested this on WPAndroid yet, but we'll probably need to see if the newly introduced serialized data class collides with an older version of data stored on a handset. To test this, we'll have to use the current version of WPAndroid with Stories, save / publish a Story, and then update the library with this PR and run it, and check if anything is wrong there. Since the new class holds a nullable field (the flattened array), I hope nothing happens, but I'll tackle that in a subsequent PR if needed.~

To test:
Just tested the following (**WPAndroid**):
1. use WPAndroid's `develop` 
2. create a Story with a couple of image slides and publish it
3. edit it, verify you can move added views (but can't resize the background image)
4. switch the `libs/stories-android` to this branch, build WPAndroid and run
5. open the Story post
6. edit it, use pinch to zoom
7. save and update the Post
8. verify the edits (including zoomed in background) were published correctly
9. open the Post once more
10. verify you can resize the background images to their original size, publish, and so on.

Using the **demo app**:
1. add images, videos, slides, etc
2. use pinch to zoom
3. verify the zoomed in portion of each image is kept as you switch between frames
4. add views
5. verify you can drag them around / resize them
6. tap on the checkmark icon on the top-right to save frames
7. verify the saved images reflect the background images zoomed in in the portions where you wanted them to be, and the added views also reflect their positioning correctly.


aforcier:
Looks good @mzorz !

I tried the test steps. Everything works fine in the demo app. On WPAndroid I'm able to pinch to zoom when editing an existing Story and all, but when saving the changes it seems to ignore the zoom changes and keep the original image. Are you able to reproduce that?

(Perhaps this will need to be resolved in a later PR/with a WPAndroid integration PR.)

mzorz:
Thanks for testing and reporting back @aforcier  🙇 . Actually I just re tested on the same handsets and emulators and could reproduce the fail you described consistently both on the demo app and WPAndroid when using only one slide 🙂 . It should work when you test several slides as per the tests - however this behavior is fixed in this essential [this part of code](https://github.com/Automattic/stories-android/pull/648/files#diff-d7b9cfc9557f5db9b24967ac4126af95c2da5cb61e3bfc30ddce176571dfbde6R305-R310), as well as refreshing the currently selected slide info when saving [here](https://github.com/Automattic/stories-android/pull/648/files#diff-91f59362418d76a8a4362ad5c6801d7e1f754165385a34df2c6244e011678a5cR2017).

So basically this issues is fixed in the follow up PR #648 👍

aforcier:
Great! In that case this PR is also :shipit:

mzorz:
Merging - this PR is [approved](https://github.com/Automattic/stories-android/pull/647#issuecomment-793303192) along with all the subsequent work which is already merged as well (see #648, #649, #651, #650)

PR #649, opened 2021-03-02
Pan / zoom background images: cropping / fit-center criteria

mzorz:
Follow up to https://github.com/Automattic/stories-android/pull/648

### EDIT: new definition
_(adding this here on top of the PR's description so it's easier to understand for future reference, what we ended up doing in this PR.)_

After back and forth and testing some APKs, and with the goal of prioritizing visual editing space and reflecting output as much as possible, we decided to go with the following definition.

After [this comment](https://github.com/Automattic/stories-android/pull/649#issuecomment-790855387) we discussed several situations and we think the following could cover all cases pretty well.

#### Cropping
- if the image being loaded matches the aspect ratio of the device screen, then load it with center-crop (as it was before these changes), so it occupies the full screen starting from the top (no parts would actually be cropped, given the matching the aspect ratio)
- if the device is taller than 9:16, just crop the bottom (showing the opaque bar)
- else, load with fit-center as per #648 

So basically the first bullet describes the situation we generally have before #648 
The second bullet means to implement `CROP_CENTER` _and_ the opaque bar as per [this](https://github.com/Automattic/stories-android/pull/649#issuecomment-789359955) (point 1) in https://github.com/Automattic/stories-android/commit/43755837b440b2d5d0f2ab73f4f12b605ff7c252 for devices taller than 9:16.
The third bullet defaults to the behavior implemented for #648 (`FIT_CENTER`).

#### Opaque bar
The opaque bar should only cover the part that is not going to make it to the exported output after 9:16 normalization. The Story Frame selector should not be fully opaque, but only semitransparent.

Definition:
Screen Shot 2021-03-05 at 09 18 59

Example:
ttttest

------
### History (this PR, originally)
Fixes the situation described in this comment https://github.com/Automattic/stories-android/pull/648#issuecomment-788506074

> On phones taller than 16:9, restrict the editable surface to 16:9 and black out the bottom section of the editor that would have been cropped out on export. (And don’t allow views to be added there/crop views that extend over the cutoff).

The working area gets cropped preemptively using the same calculations we use for export.
For example on a Pixel 3 emulator, which screen is 1440x2792, it gets limited to 1440x2560.

**Before the change** (tested on a Pixel 3 emulator):
Screen Shot 2021-03-02 at 08 27 19

**After the change:**
Screen Shot 2021-03-02 at 08 27 29


aforcier:
@mzorz I gave this a try but I'm seeing a few issues:

1. It doesn't look like the editing surface is getting blacked out? On a Pixel 5, the image still reaches down to the bottom of the screen, even though the bottom few cm will be cut on export
2. I see that added views respect an invisible lower boundary now and can't be dragged below it. That's good, however it's possible to pinch to zoom and rotate such that some of the added text extends below the invisible line and will be cropped. This goes with the above point - if we black out the bottom part of the screen, we can also make any part of the text added to that space be hidden.
3. Related to the above point - on a tall screen it can now be pretty difficult to delete an added view, since the delete button can be outside the lower bounds for moving the view. We should probably set the lower bound of the delete button to the lower bound of the actual surface.

mzorz:
Good points @aforcier ! I had to dig a bit into past decisions as it's one of those thorny UX areas where modifying behavior here makes things worse there 😅, let me share my thoughts (this is only to reaffirm what you said but I needed to get back and check history, and I'm completing information here for our future selves as well):

#### On point 1
> 1. It doesn't look like the editing surface is getting blacked out? On a Pixel 5, the image still reaches down to the bottom of the screen, even though the bottom few cm will be cut on export

You're right - I thought it would suffice to only limit the interaction but indeed I should have mentioned what I thought. It would only get blacked out (letterboxed) in the preview / editing mode when the media size doesn't fit the device screen size as per #648. On the Pixel 5 (or more generally taller than 9:16 screens) it will be cut on export.

I did some digging to make sure I was on the same page, copying a few notes from there here `pbD5rk-id-p2` for completeness:
> We were focusing our convo there on screens larger (taller) than 9:16 so, cropping at the bottom would suffice, and it seemed to fit our use case well given that part of the screen is occupied with our own bottom slide selector.

[...]

And from `pbD5rk-gs-p2#comment-247` (bold marks mine):

> Alternative approach – Always export 9:16 media
> To reduce the amount of variability with hiding parts of the media, one thing we might want to consider is having the mobile editors always output a media size that is 9:16, regardless of the device screen ratio.
> Here’s how that might look. **I don’t believe it would require any UI changes if we crop from the bottom “safe area” where the slide thumbnails are.**

[...]

In fact, there's a hint in that conversation that we could make the bottom stripe 100% opaque to make it obvious as point 1 seems to suggest we should do, but we never decided firmly on that:

> To make it extra obvious to the user, we could make the opacity on the bottom bar 100% black on devices taller that 9:16

Ok so back to your original comment (forgive the delay on that); I went ahead and as such implemented this part:
```...and black out the bottom section of the editor that would have been cropped out on export.```  in commit 43755837. In that commit, we are adding a mechanism for both setting the `StoryFrameSelector`'s background to either opaque black or semi transparent, as well as adding a new opaque View only for devices with a screen aspect ratio taller than 9:16. We also take into account we should hide the bar when coming back to capture mode, to offer the full screen real estate for camera live preview.

Unfortunately, accounting for this while seems to work well in terms of setting expectations re: final output, it leads to mixed results that are not great in terms of the visual appearance:
- if the calculated excess area (which would be cut off after normalizing to 9:16) is not tall enough to reach the StoryFrameSelector, then it would leave a visible area in between the bottom of the Selector and the top of the new opaque bar
- if the calculated excess area height is less than the navigation bar's height, the Android navigation controls can be seen as unaligned
- if the calculated excess area is taller than the StoryFrameSelector's top border, the StoryFrameSelector looks unaligned

Here's a few screenshots covering some of these issues

Screen Shot 2021-03-03 at 11 17 36
Screen Shot 2021-03-03 at 11 19 44

We could take a few compromises but in the end it would mean that the visuals would not exactly reflect the output, which is what I understand was the desired effect we were searching for from our conversations elsewhere.

#### On point 2
> I see that added views respect an invisible lower boundary now and can't be dragged below it. That's good, however it's possible to pinch to zoom and rotate such that some of the added text extends below the invisible line and will be cropped. This goes with the above point - if we black out the bottom part of the screen, we can also make any part of the text added to that space be hidden.

The possibility to pinch and rotate makes things harder in terms of detecting collision with the outer bounds. It makes the draggable area so big (I mean, the containing square / rectangular view) that it's difficult to even move it around given it easily collides with the edges. In fact this is I would say a known issue - In this sense I think I would allow people to drag views around, just hinting them visually that they are going off the limits of the screen but still allowing free movement as much as possible, to avoid that unpleasant feeling of not being able to drag things where you want.

> if we black out the bottom part of the screen, we can also make any part of the text added to that space be hidden.

That's right. For this and what I mentioned above, I'm inclined to think we should probably remove the code in commit 19066d57106e4fb5d62adb055a6a0ff514e1ebc2 in this PR, and instead just leave users with the visual hint (IFF we resolve the visual problems described above in the different screenshots).

#### On point 3
> Related to the above point - on a tall screen it can now be pretty difficult to delete an added view, since the delete button can be outside the lower bounds for moving the view. We should probably set the lower bound of the delete button to the lower bound of the actual surface.

Again - I would probably remove the first commit in this PR and stick to visual hints, but allow users freedom of movements more. 

Let me know your thoughts @aforcier ! 🙇

mzorz:
Note: realized moving added views around would always be on top z-index wise so, refactored and simplified things a bit by moving  the opaque bottom bar to ComposeLoopFrameActivity's container in https://github.com/Automattic/stories-android/pull/649/commits/6afe6122ffdc3597c49068b4169ed24cd095a5ad, which makes z-index so much easier to handle (i.e. no actual handling, the default makes the bottom opaque bar to appear at the desired level, while added views happen at a lower level, always being covered by the opaque bar when present)

aforcier:
Gave this a try @mzorz , you're right it is close to what we ended up deciding in the call. Some screenshots, also to illustrate for @megsfulton:

(Editor preview on the left, exported image on the right)

Pixel 3 (A bit taller than 9:16, but not much):

649-pixel3

Pixel 5 (Significantly taller than 9:16):

649-pixel5

### Bugs:

* In both cases, it looks like the blacked out area is larger than it needs to be, since the exported photo had some text that was cut off in the preview. In the Pixel 3's case it's probably fine (the alternative would be to make the blacked out area start around the middle of the thumbnail bar, which maybe looks weird).
* The delete button is unreachable in the Pixel 5's case (and hard to get to on the Pixel 3 as well). I think in your comment you're proposing resolving this by removing the limits and just relying on the blacked our areas, which I think makes sense.

### Implementation concerns:

The blacked out area on the Pixel 5 is quite dramatic (though part of this seems like a bug and it doesn't actually need to extend beyond the top of the text editor button). Megs' suggestion to also black out from the top seems like a good idea here.

mzorz:
Thanks for testing this @aforcier! 🙇 

> In both cases, it looks like the blacked out area is larger than it needs to be, since the exported photo had some text that was cut off in the preview. In the Pixel 3's case it's probably fine (the alternative would be to make the blacked out area start around the middle of the thumbnail bar, which maybe looks weird).

I think the Pixel 3 case is that the background color of the StoryFrameSelector is black, so it gives a sense that the top part of it is the actual cropping limit, when the actual cropping limit is somewhere in the middle-to-bottom of the actual StoryFrameSelector, as in this picture here:
Screen Shot 2021-03-03 at 11 19 44
In this image I left the background coloring to be semitransparent on purpose to describe the problems better, but the actual code in the PR would make it black as per your tests.

I'll look into the Pixel 5 case and come back here.

> The delete button is unreachable in the Pixel 5's case (and hard to get to on the Pixel 3 as well). I think in your comment you're proposing resolving this by removing the limits and just relying on the blacked our areas, which I think makes sense.

Do you think this is the same for previous APKs (i.e. `develop`)?  or does this situation seem to be made worse in this PR in particular?

mzorz:
> I'll look into the Pixel 5 case and come back here.

I was able to reproduce on an emulator with a custom hardware profile, will get back here to update 👍

megsfulton:
Gave this a try on a Pixel 3. 

Ran into an issue where text objects can't be deleted because the trash icon is in the uneditable area so the text can't be dragged to it. 

I uploaded an image that was originally 9:16 and it was exported with a black bar across the top. For a case like this I expected to be able to move the image around in the editable 9:16 space so the full image was shown. Or to have the image placed in the 9:16 space so the bottom isn't getting cropped out. 

Editor
Screenshot_20210304-105721

Export
wp_story1614884375365_0

mzorz:
> The blacked out area on the Pixel 5 is quite dramatic (though part of this seems like a bug and it doesn't actually need to extend beyond the top of the text editor button).

This is fixed in f4745ab6, it was indeed a bug 👍

mzorz:
> The delete button is unreachable in the Pixel 5's case (and hard to get to on the Pixel 3 as well). I think in your comment you're proposing resolving this by removing the limits and just relying on the blacked our areas, which I think makes sense.

reverted in b90e3e1e0da0030d1ee61d892f71092701fbfa55

mzorz:
Heads up - changed the PR description and title to reflect our discussions elsewhere after this [comment](https://github.com/Automattic/stories-android/pull/649#issuecomment-790855387) and the aim of this PR. Going to work on that now.

mzorz:
Set this one back to "ready for review" @aforcier - this PR only adds the opaque bar on the visual editing area when needed (i.e. for screens taller than 9:16), and precedes all the work in #651 (i.e #651 is based on top of this PR)

aforcier:
This one's good to go! :shipit:

PR #650, opened 2021-03-02
Pan / zoom improvements: using glide-transformations library to add blurried background

mzorz:
This builds on top of #651 (last step to pan / zoom improvements)

>  On all phones, allow pinch to zoom, and when either height or width is now larger than the image bounds, **show a blur which will be included in the exported image.**

I first tried to reuse PhotoEditor's [PhotoFilter](https://github.com/burhanrashid52/PhotoEditor/blob/master/photoeditor/src/main/java/ja/burhanrashid52/photoeditor/PhotoFilter.java), which is based on Android's Media package [EffectFactory](https://developer.android.com/reference/android/media/effect/EffectFactory) but it doesn't have a Blur effect builtin so, we can't rely on that (see https://github.com/burhanrashid52/PhotoEditor/wiki/Filter-Effect).

After checking a bit I decided to fo with [`glide-transformations`](https://github.com/wasabeef/glide-transformations) library which had a handy blur transformation for Glide.

Once that's sorted out we need to decide on an architecture to comply with two things:
- [x] each time the Story frame is changed, we need to present the blurry background, the source image, and the Added Views on top.
- [x] the resulting output file should also have the blurried background in it

We're taking advantage of the pre-existing [`OnImageChangeListener`](https://github.com/Automattic/stories-android/blob/df17e8c19d4e5584a7002e042adef4a5d4c32f9c/photoeditor/src/main/java/com/automattic/photoeditor/views/background/fixed/BackgroundImageView.kt#L38-L40) interface, which has one callback `OnBitmapLoaded` which gets called on several / all overridden setter methods that may have a visual impact on the underlying ImageView.

With this, each time an image background is set, we also make a call to draw the transformed /  blurred image on the background (deeper z-level).

**Example**
For these 2 original background images:
512x512 PNG Landscape Texture - Rocky Stream
Icon_Bird_512x512

This is the resulting output on a Pixel 3 XL (see it's normalized to 9:16):
wp_story1614805120565_2
wp_story1614805120758_1

Notes:
~- Still need to figure out changes to FrameSaveManager, given we now need to paint both the blurred background and the background on the ghostPhotoEditor and we may need to wait and chain these together with Glide's [listener](https://bumptech.github.io/glide/javadocs/440/com/bumptech/glide/request/RequestListener.html#onResourceReady).~ this wasn't needed, just loading both is ok given it's z-index will paint them correctly, and also we're using sequential calls within our coroutines (which means async callback chaining wasn't going to be possible nor needed anyway)
- There's a small (but noticeable) delay on the Samsung J2 I'm using for testing after the main image is shown and before the blurred background is shown.

### DIFF
--- a/photoeditor/build.gradle
+++ b/photoeditor/build.gradle
@@ -54,6 +54,8 @@ dependencies {
     implementation 'com.github.bumptech.glide:glide:4.10.0'
     kapt 'com.github.bumptech.glide:compiler:4.10.0'
 
+    implementation 'jp.wasabeef:glide-transformations:4.3.0'
+    
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
 
     implementation project(path: ':mp4compose')
--- a/photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
@@ -79,6 +79,9 @@ class PhotoEditorView : RelativeLayout {
     val source: ImageView
         get() = backgroundImage
 
+    val sourceBlurredBkg: ImageView
+        get() = backgroundImageBlurred
+
     val brush: BrushDrawingView
         get() = brushDrawingView
 
@@ -118,6 +121,10 @@ class PhotoEditorView : RelativeLayout {
     override fun onAttachedToWindow() {
         super.onAttachedToWindow()
         attachedToWindow = true
+    }
+
+    fun onComposerDestroyed() {
+        attachedToWindow = false
     }
 
     @SuppressLint("Recycle")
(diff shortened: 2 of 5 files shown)


aforcier:
Working great @mzorz ! Since this should be the last in the PR stack, I also built WPAndroid with this branch and tested editing and publishing on a variety of devices and for different file types/sources, everything worked smoothly (including [this issue](https://github.com/Automattic/stories-android/pull/647#issuecomment-792828854) which you mentioned had been fixed since I saw it in the original base PR).

With the discussion around the remaining issues Megs reported + https://github.com/Automattic/stories-android/issues/653 filed, it looks like the feature is good to go!

:shipit: - I'll leave it to you to collapse the stack and get the WPAndroid PR ready.

mzorz:
Thank you for all the testing and reviews @aforcier ! Will now proceed to merge and prepare the WPAndroid PR updating the stories library commit hash 👍

PR #651, opened 2021-03-05
Pinch to zoom / implemented Glide loading with FitCenter or CenterCrop

mzorz:
Builds on top of #649.

### Description
In the spirit of aiming at prioritizing visual editing space, and then reflecting output (which we normalize at 9:16),  this PR implements the following logic:

- if the image being loaded matches the aspect ratio of the device screen (or has close one), then load it with center-crop (as it was before these changes), so it occupies the full screen starting from the top (no parts would actually be cropped, given the matching aspect ratio)
- if the device is taller than 9:16, just crop the bottom (showing an opaque bar to reflect what the export is going to look like)
- else, load with fit-center

With this, we need a previous step in which we determine the source image size. This was a bit tricky but then realized we won't ever export something bigger than the screen size (given the edits) so I started with loading the image using the `override()` Glide builder method with using the values for `screenWidth`, `screenHeight`.

### Logic

The main problem was to identify the source image’s dimensions first without wasting resources or having to pre-load it, doing it with Glide is complicated since it optimizes loading for the target view, and obtaining the size without loading is not a use case they support. Doing it as per the Android docs is straightforward for when the image exists as a File but not when it’s being fed to the app from a third party such as an input stream, and there’s other corner cases. Given we won’t ever output an image bigger than the device screen size, setting that as the upper limit for the target view seemed to work (otherwise I ran into memory and performance issues).

Then the decision process compares the image size once adapted to the target view with Glide, and takes it from there (code can be seen [here](https://github.com/Automattic/stories-android/pull/651/files#diff-91f59362418d76a8a4362ad5c6801d7e1f754165385a34df2c6244e011678a5cR2017-R2075)) I was trying to resize the PhotoView to match the output but it made things worse, so I had to investigate further and decided to use Glide’s override() method to target the output size

The process works in 2 steps: step one (`loadImageWithGlideToPrepare()`) is to decide what’s the best output for the scenario (crop_center for images that resemble the aspect ratio of the available screen size, fit_start/fit_center for taller than 9:16 devices with portrait images, and then fit_center for every other case), and the second step (`loadImageWithGlideToDraw()`) is to actually draw on the target view

### Test cases
**CASE A: 9:16 device**
Using a device screen example of 1080x1920 (i.e. Pixel), 9:16 or Samsung J2 549x960 _(could be anything as long as the set of images used comply with aspect ratio)_
Observe the following:
- the opaque bar is hidden from the visual editing area, since the exported image will be the same size of the screen
- image A same / similar aspect ratio --> center_crop, most of the image is seen
- image B portrait --> FIT_CENTER
- image C landscape --> FIT_CENTER

**CASE B: taller than 9:16 device**
Using a device screen example of 1080x2340 9:19.5 approx
Observe the following:
- the opaque bar is shown on the visual editing area to show what's getting cropped in the output
- image A equal size --> center_crop, most of the image is seen
- image B portrait --> if wider than screen, FIT_START
- image C portrait --> if narrower than screen, FIT_CENTER (so it doesn't look uncentered)
- image D landscape --> FIT_CENTER

**CASE C: shorter device**
Using a device screen example of  device screen example: 1440x2960 (i.e. Samsung S8+), 800x1280 (3:5), Nexus 4 (768x1280, 3:5)	
Observe the following:
- the opaque bar is hidden (given the device screen is not taller than 9:16)
- image A equal size --> center_crop, most of the image is seen
- image B portrait --> FIT_CENTER
- image D landscape --> FIT_CENTER


aforcier:
I ran through all the test cases and everything seems to be working great 🎉 I haven't had a chance to dig into the code yet, will do that soon 👍

stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
```diff
@@ -2003,11 +2006,105 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
...
+        val model = (frame.source as? FileBackgroundSource)?.file
+                ?: (frame.source as UriBackgroundSource).contentUri
+
+        CoroutineScope(Dispatchers.IO).launch {
+            val futureTarget = Glide.with(this@ComposeLoopFrameActivity)
+                    .asDrawable()
+                    .load(model)
+                    .fitCenter() // we use fitCenter at first (instead of cropping) so we don't lose any information
+                    .submit(screenWidth, screenHeight) // we're not going to export images greater than the screen size
+            val drawable = futureTarget.get()
```
aforcier, line 2025: This line generates a concerning warning in AS: `Inappropriate blocking method call `.

I'm not very well versed in Coroutines but by inspection I think this is a false positive - we're launching this within a `CoroutineScope` wrapping the `IO` dispatcher, so we should not be blocking the main thread. I'm just not sure if this is an improper call within the `IO` context and we should be offloading this blocking call in some other manner, but it seems more likely that the linter doesn't really understand the context around this blocking call. I thought I'd raise it anyway though, do you think there's any cause for concern here @mzorz ?

mzorz: Good question @aforcier !

It's safe to call a [blocking operation](https://developer.android.com/kotlin/coroutines/coroutines-adv#manage) from a coroutine - that said, using [`launch` to start](https://developer.android.com/kotlin/coroutines/coroutines-adv#start) the coroutine is good for a fire and forget operation such as this one where we want to fetch the image first (using Dispatchers.IO) and then switch context to draw on the UI using `withContext(Dispatchers.Main)` in it.

So yes I think you're right and the linter is mistakingly taking this as a positive when it isn't.

aforcier: Makes sense, just wanted to double-check, thanks @mzorz 👍

aforcier:
I gave this a deeper pass today and everything looks good to me 👍 Really nice work on all this Mario 🎉 

:shipit:

PR #655, opened 2021-04-09
Fix: make sure to run onStoryFrameSelected in onLoadFromIntent

mzorz:
I wasn't able to pinpoint exactly when this started happening, but when you load an existing Story the first frame won't be refreshed. We're now forcing that refresh with this PR. so first frame is loaded when loading from intent

To test:
1. create a Story on WPAndroid
2. try editing it
3. observe the first frame appears correctly


aforcier:
Confirmed it's working correctly - :shipit:

PR #660, opened 2021-04-20
Always reorder AddedViewList according to z-index order in getter

### LINKED ISSUE #534: Bug: z index / elevation of objects in the story editor is different in the output
> I took the first screenshot right before publishing, and the second is a screenshot of the story viewer. The emoji should be behind the text, but their ordering got mixed up on publish.

Screen Shot 2020-09-16 at 5 21 33 PM

mzorz:
Fix #534 

This PR introduces a change to make sure to always keep an `AddedViewList` that is up to date with the order in which added views are appearing on screen.

At first, I explored overriding [`getChildDrawingOrder()`](https://developer.android.com/reference/android/view/ViewGroup#getChildDrawingOrder(int,%20int)) to keep a list of z-index ordered views at the `PhotoEditorView` level but it turned out not to be necessary, given the order in which the views are returned by the `children` property in `ViewGroup` is already z-index ordered.

Hence, I went ahead and only added a helper method in `PhotoEditor.getViewsAdded()` called `reorderAddedViewListAccordingToZIndex()`, which will just iterate over the actual ViewGroup's `children` collection and find its corresponding `AddedView` object from its referenced native Android `View` and re-build the `AddedViewList` accordingly before returning it to the saving / re-drawing function.

### Before

https://user-images.githubusercontent.com/6597771/115372443-45ee8c00-a1a1-11eb-9781-b0a4a92ad17a.mp4

### After

https://user-images.githubusercontent.com/6597771/115372406-38d19d00-a1a1-11eb-8903-a1b6f4e026e8.mp4

To test:

1. run the demo app, tap on + to add a new slide
2. take a picture or choose from the picker
3. add one view (Text), make sure to remember it was the first view you added
4. add another view (Text), make sure to remember it was the second one you added
5. add a new slide
6. select the previous slide, observe the order in which views are drawn
7. tap on the first view and make it overlap, so you know the first view you added is now on the front
8. switch slides, then switch back to the slide where you added views
9. observe the z-index order is kept (i.e. the last touched added view is drawn on top)

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
@@ -1046,7 +1046,24 @@ class PhotoEditor private constructor(builder: Builder) :
     }
 
     fun getViewsAdded(): AddedViewList {
+        // always make sure to return a freshly z-index-ordered AddedViewList
+        reorderAddedViewListAccordingToZIndex()
         return addedViews
+    }
+
+    private fun reorderAddedViewListAccordingToZIndex() {
+        val reorderedAddedViewList = AddedViewList()
+        for (oneView in parentView.zIndexOrderedAddedViews) {
+            when (oneView.tag) {
+                TEXT, EMOJI, ViewType.STICKER_ANIMATED -> {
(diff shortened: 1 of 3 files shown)


aforcier:
Works great! Confirmed the steps, and also checked the actual export :shipit:

PR #666, opened 2021-04-26
Update opaque bar height in OnApplyWindowInsetsListener

### LINKED ISSUE #664: Bottom opaque bar on editing: it only draws itself when reordering slides
When editing an existing Story, the [bottom opaque bar](https://github.com/Automattic/stories-android/pull/651) which is drawn on purpose to reflect 9:16 exports is only shown when a re-order action is about to take place (tap and hold on a frame on the Story Frame selector).

**What should happen**
The bottom It should be drawn right away upon having the screen rendered.

Tested on: Pixel 4a, Android 10.

See video:

https://user-images.githubusercontent.com/6597771/116056993-bf302800-a654-11eb-85bd-05156afd10a2.mp4

In the video you can see the frame selector is transparent, and it only draws itself once a reorder action is about to take place (tap and hold on a Story frame)

mzorz:
Fix #664 

The calculated opaque bar height was not being correctly applied after having changed so this PR fixes that by making sure the height is set on the opaque bar view right after calculating it in the `OnApplyWindowInsetsListener`.

Then, instead of setting the height at a later moment when the bar needs to be shown, we're only setting the `visibility` attribute to handle show/hide behavior, making it more precise.

To test:
Using a Pixel 4a (device with a screen taller than 9:16)
1. On WPAndroid: create a Story, add a slide, publish it
2. open the post and edit the Story
3. observe the opaque bar is correctly drawn

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -356,6 +356,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
             workingAreaRect = calculateWorkingArea()
             photoEditor.updateWorkAreaRect(workingAreaRect)
             bottomOpaqueBarHeight = preCalculateOpaqueBarHeight()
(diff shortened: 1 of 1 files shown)


aforcier:
:shipit:

PR #669, opened 2021-04-27
Fix: add touchlistener by default on newly added text added views

mzorz:
In https://github.com/Automattic/stories-android/pull/663 we added an optional addition of the touch listeners to added view, with the parameter being `false` by default. I omitted making it explicit `true` when adding new text to a Slide 🤦 

To test:
1. create a Story
2. add a slide
3. add text
4. try to drag it around, touch to edit, zoom, etc. and verify it works


aforcier:
🤦 my bad too, I guess I didn't check moving added views, though I do remember moving text around - must have been while editing an existing slide.

:shipit:

PR #670, opened 2021-04-28
Only set opaque bar height if the calculated result is greater than 0

mzorz:
Fix an issue introduced in https://github.com/Automattic/stories-android/pull/666

When the phone screen is exactly 9:16, the calculated opaque bar height is 0, but setting the LayoutParams' height to zero will indicate it to use all the available space (similar to `match_parent`).

Hence, we should just not set it when the result is zero.

To test:
1. Use a device with a 9:16 screen (I used the original Pixel)
2. create a Story
3. observe the background image for a slide is shown (otherwise without this patch, it'd turn black given the opaque bar height is set to cover the screen)


aforcier (approved):
Was able to reproduce on a Pixel emulator. Though starting a story from camera capture didn't present the issue, I had to either create from gallery, or add from camera and then add another slide and switch slides for the issue to appear.

Confirmed the fix. :shipit:

PR #671, opened 2021-04-28
Fixes for Retry button

mzorz:
Fixes #577 

- make the StoryFrameSelector INVISIBLE instead of GONE so it occupies Screen space and doesn't make the Retry button fall below the navigation buttons. (https://github.com/Automattic/stories-android/pull/671/commits/6b22a5a73a9bb2de736a1abef667743dafd5c9ca)
- make the background plain white so it's clearer when it's behind the block-screen schrim. (https://github.com/Automattic/stories-android/pull/671/commits/6b22a5a73a9bb2de736a1abef667743dafd5c9ca)

Plus:
- ~Only allow slide deletion mode toggling when not in error handling mode (d009dd6c)~
- Show/hide retry button when toggling delete mode, otherwise they overlap (31d673e4 and e3c505cb)
- Check Story index is valid in the StoryRepository when attempting to load from an error notification cd35c1ce (there may be a case when the Story is errored and then an unrelated crash in the app will still allow the notification to persist so, the storyindex won't be valid anymore).

This is the crash that is fixed here:

```
2021-04-29 14:02:40.614 4432-4432/org.wordpress.android D/AndroidRuntime: Shutting down VM
2021-04-29 14:02:40.615 4432-4432/org.wordpress.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.android, PID: 4432
    java.lang.RuntimeException: Unable to start activity ComponentInfo{org.wordpress.android/org.wordpress.android.ui.stories.StoryComposerActivity}: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3431)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3595)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7660)
        at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #673, opened 2021-05-01
Fixes for Retry button (part 2)

### LINKED ISSUE #672: Crash when retrying saving an errored slide more than once
```
2021-04-29 22:27:05.470 23371-23371/org.wordpress.android D/AndroidRuntime: Shutting down VM
2021-04-29 22:27:05.473 615-615/? E/SELinux: avc:  denied  { find } for pid=17308 uid=10246 name=tethering scontext=u:r:secure_ui_service_app:s0:c246,c256,c512,c768 tcontext=u:object_r:tethering_service:s0 tclass=service_manager permissive=0
2021-04-29 22:27:05.475 23371-23371/org.wordpress.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.android, PID: 23371
    java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at java.util.ArrayList.get(ArrayList.java:437)
        at com.wordpress.stories.compose.ComposeLoopFrameActivity$saveStoryPostHook$1.run(ComposeLoopFrameActivity.kt:1202)
        at com.wordpress.stories.compose.SaveButton$showSavedAnimation$1.run(SaveButton.kt:52)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.app.ActivityThread.main(ActivityThread.java:7660)


```

**Steps to reproduce:**
1. apply this [patch](https://gist.github.com/mzorz/41d6f80f2395105fff6da6328b190ff5) to force an error on video slides
2. create a Story with 1 video slide at least
3. observe it throws an error (see error notification and the snackbar on My Sites screen if on WPandroid)
4. tap on the notification body, on the `MANAGE` quick action or on the `MANAGE` snackbar action 
5. observe the error resolution mode in Composer Activity appears
6. tap the errored frame, tap on RETRY, and wait for it to error again
7. tap on RETRY once more, observe the app crashes with the logs shown above.

mzorz:
This builds on top of #671 

Fixes #672 and more

This fixes some other issues with retries:

- if an error occurs and the user taps on the Retry button, after a new attempt to save the addedviews will have been gone away, because we need to detach them from the real PhotoEditorView instance to use the ghost PhotoEditor instance for saving in the FrameSaveService, and then re-add these views to the PhotoEditorView on the real screen using the UI thread. For this, we're now passing the `isRetry` flag as a parameter in `saveStory()` in the Service, and reconverting its meaning to `reattachAddedViewsAfterSaving` in the FrameSaveManager. (https://github.com/Automattic/stories-android/pull/673/commits/80d811c247e5bce2bf7ded13487e61355eafc904). This in the end would be worsened after a couple of retries and the crash in #672 would be run into (so, this commit  80d811c247e5bce2bf7ded13487e61355eafc904 fixes that)
- in some strange cases, access to AddedViews could be done concurrently from different coroutines so, marked the methods `@Synchronized`  now (https://github.com/Automattic/stories-android/pull/673/commits/a49a6f39dcfb0228b6b4b34fc0e0e187acd557da)

To test:
1. apply this [patch](https://gist.github.com/mzorz/41d6f80f2395105fff6da6328b190ff5) to force an error when saving, and create a Story
2. observe the error Notification and snackbar appear. Tap on `MANAGE`.
3. Retry the failed slides
4. observe while retrying, the added views may flicker (so these are released and re-added later so these can be properly saved), but they are no longer lost
5. the patch will make things fail every time, but no other side effects should be seen.

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
@@ -1060,12 +1060,14 @@ class PhotoEditor private constructor(builder: Builder) :
         return (addedViews.size > 0)
     }
 
+    @Synchronized
     fun getViewsAdded(): AddedViewList {
         // always make sure to return a freshly z-index-ordered AddedViewList
         reorderAddedViewListAccordingToZIndex()
         return addedViews
     }
 
+    @Synchronized
     private fun reorderAddedViewListAccordingToZIndex() {
         val reorderedAddedViewList = AddedViewList()
         for (oneView in parentView.zIndexOrderedAddedViews) {
--- a/stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveManager.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/frame/FrameSaveManager.kt
@@ -66,7 +66,8 @@ class FrameSaveManager(
 
     suspend fun saveStory(
         context: Context,
-        frames: List<StoryFrameItem>
+        frames: List<StoryFrameItem>,
+        isRetry: Boolean
     ): List<File> {
         // calling the listener here so the progress notification initializes itself properly and
         // shows really how many Story frame pages we're going to save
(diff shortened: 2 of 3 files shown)


aforcier (approved):
Looks good, :shipit: !

I was running into an odd apparent framework crash while testing this which I shared with you, but I don't see how it would be related to these changes. I'll approve this but leave it to you to merge once you've taken a look at the crash too.

PR #674, opened 2021-05-03
Temporarily remove fetchstyle & configure plugins

oguzkocer:
We have to temporarily disable installable builds because configure plugin is necessary to build them.


jkmassel (approved):
Lots of comments – straightforward enough!

Issue #678, opened 2021-05-05
Limited Story Editing dialog appears when editing right after saving


mzorz:
It occurred to me many times that, after saving a Story on WPAndroid and then tapping on the Post to edit -> tap twice on the Story block to select and open on the Storycomposer, for some reason some of the slides wouldn't load (generally, never the case for the first slide but for slides on the right). Sometimes they don't even appear on the frame selector, sometimes they do but seemingly no information could be found in Prefs or related to background image.

If you try several times (exit, tap again to edit, etc.) at some point it looks like the information is available and now you can fully edit the Story. This seems to be consistent, in the sense that subsequent tries will always open correctly.

mzorz:
Closing via https://github.com/Automattic/stories-android/pull/681

PR #562, opened 2020-10-09
Tap to delete slide

aforcier:
Closes #423.

Deleting a slide is now done by tapping the current slide in the frame selector, and I removed the overflow menu we used to keep this functionality in.

tap-to-delete-slide

### To test:
1. Tapping a slide should hide most of the UI and show the delete button
2. Tapping on another slide, or the same slide again, dismisses the delete button
3. Tapping anywhere else on the UI should dismiss the delete button and handle the tap (e.g. tapping on an added view will open the editor)
4. Adding a new slide while the delete button is visible should dismiss the delete button

PR #681, opened 2021-05-06
Move deleteCapturedMedia() to OK listener for frame removal dialog

### LINKED ISSUE #680: Video slide not playing when opened after a while of being recorded
Observed first while testing #679 

On the Pixel 4a, I've observed a video recorded with the CameraX camera is only playable for a few moments, but after publishing it only works the first few times. At some point it just keeps showing the "loading" spinning wheel.

This only happens with locally recorded videos, but it works alright when you select videos from the media picker (i.e. available through other apps)

mzorz:
Fix #680, #678

### Getting to the cause
What was happening here is that, in the case we left the Story composer with a locally captured file as the last Story frame visited, we were actually deleting the original captured file. Hence, the next time we visited it, it would be lacking the background file, leading to 2 problems:
1. "Limited Story Editing" dialog appearing, because we couldn't [validate we had all the background files](https://github.com/wordpress-mobile/WordPress-Android/blob/70ce7de8817020d29768074bb408688436320220/WordPress/src/main/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCase.kt#L46-L61) for this Story (#678) - _a note on the issue listed there, when I wrote the issue I described what I thought was happening but it's not exactly that it would become available later (I haven't been able to reproduce it since I've been able to wrap my head around this one)_
2. Video files not being found when the video slide was captured with the phone so, getting the spinning wheel altogether. (#680)

### History
This was fist introduced in the early days when we didn't even have a slide selector
https://github.com/Automattic/stories-android/commit/ddab8bedee4ed0c020542d5103327e51f77041b6

There's a change introduced in this commit https://github.com/Automattic/stories-android/commit/4145288e3148aa4bdd21309a7643206682a5b363 where we were setting the `currentOriginalCapturedFile` to be the background file path for the selected slide. Then, if the user decided to discard a frame we would be able to delete it's backing captured file.

Later on, we moved the "discard frame?" Dialog code elsewhere when implementing the "tap to delete" behavior (see https://github.com/Automattic/stories-android/commit/427f1bc3ddb3c3252520010d2d7fd5e3d5d4b51e, PR https://github.com/Automattic/stories-android/pull/562). I think in order to keep the same behavior as before, we should be also able to delete the original captured media file there, as it was being done previously.

### Solution
As said above, this PR just moves the calls to `deleteCapturedMedia()` to the OK listener of the "Delete frame" confirmation dialog. This way, we'll make sure to only discard a captured image / video File for which a slide won't exist anymore.

### Next
Idea: it may be better to just leave any captured file live and let the user clean it up later if they wanted to do so. With that, we would also avoid bugs like this and would make the code simpler.

### To test
1. create a Story using the media picker
2. edit it from Gutenberg
3. add a new video slide using capture mode (this will create a local File)
4. observe you can switch slides and see the background image ok for all slides (including the last one)
5. let the last slide (the one right before the video added in step 3) be the one you're looking at before exiting
6. exit the editor (tap back and discard changes). At this point in time, we'd delete the file in `develop` as of now.
7. edit the Story again
8. observe you can tap on the last slide and it will be showing up correctly, and also no "Limited story editing" dialog shows up.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -613,9 +613,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
     private fun setupStoryViewModelObservers() {
         storyViewModel.uiState.observe(this, Observer {
             // if no frames in Story, finish
-            // note momentarily there will be times when this LiveData is triggered while permissions are
-            // being requested so, don't proceed if that is the case
-            deleteCaptureMediaAndFinishWhenEmptyStory()
+            finishWhenEmptyStory()
         })
 
         storyViewModel.onSelectedFrameIndex.observe(this, Observer { selectedFrameIndexChange ->
@@ -631,11 +629,9 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
         })
     }
 
-    private fun deleteCaptureMediaAndFinishWhenEmptyStory() {
+    private fun finishWhenEmptyStory() {
         if (storyViewModel.getCurrentStorySize() == 0 &&
                 firstIntentLoaded && !permissionsRequestForCameraInProgress) {
-            // finally, delete the captured media
-            deleteCapturedMedia()
             finish()
         }
     }
@@ -892,7 +888,7 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
             } else if (intent.hasExtra(requestCodes.EXTRA_LAUNCH_WPSTORIES_MEDIA_PICKER_REQUESTED)) {
                 // if coming from the PHOTO_PICKER with a cancel action, and we launched with
                 // EXTRA_LAUNCH_WPSTORIES_MEDIA_PICKER_REQUESTED to start the Story with, we should cancel.
-                deleteCaptureMediaAndFinishWhenEmptyStory()
+                finishWhenEmptyStory()
             }
         }
     }
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -617,8 +617,6 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
             // being requested so, don't proceed if that is the case
             if (storyViewModel.getCurrentStorySize() == 0 &&
                     firstIntentLoaded && !permissionsRequestForCameraInProgress) {
-                // finally, delete the captured media
-                deleteCapturedMedia()
                 finish()
             }
         })
@@ -1085,6 +1083,8 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
                     listener = object : FrameSaveErrorDialogOk {
                         override fun OnOkClicked(dialog: DialogFragment) {
                             dialog.dismiss()
+                            // first of all, delete the backing captured media for this slide
+                            deleteCapturedMedia()
                             if (storyViewModel.getCurrentStorySize() == 1) {
                                 // discard the whole story
                                 safelyDiscardCurrentStoryAndCleanUpIntent()


aforcier:
@mzorz I think something may have gone wrong with the merge conflict resolution commit you pushed - the original removal of a `deleteCapturedMedia()` call (from 02321de) is no longer in the overall PR diff. I don't think that was intentional?

mzorz:
> I think something may have gone wrong with the merge conflict resolution commit you pushed - the original removal of a deleteCapturedMedia() call (from 02321de) is no longer in the overall PR diff. I don't think that was intentional?

Sorry for the miss @aforcier (and nice catch!) Hopefully that's fixed with this change now 71b308b3

aforcier:
@mzorz it's probably unrelated to this change, but I'm not able to test this in current WPAndroid `develop`. I get this crash whenever I try to select media from the media picker:

```
2021-05-12 11:32:15.405 17958-17958/org.wordpress.android.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.android.beta, PID: 17958
    java.lang.RuntimeException: Unable to start activity ComponentInfo{org.wordpress.android.beta/org.wordpress.android.ui.stories.StoryComposerActivity}: kotlin.UninitializedPropertyAccessException: lateinit property editorMediaListener has not been initialized
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3431)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3595)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.app.ActivityThread.main(ActivityThread.java:7660)
     Caused by: kotlin.UninitializedPropertyAccessException: lateinit property editorMediaListener has not been initialized
        at org.wordpress.android.ui.stories.media.StoryEditorMedia.advertiseImageOptimisationAndAddMedia(StoryEditorMedia.kt:61)
        at org.wordpress.android.ui.stories.media.StoryEditorMedia.onPhotoPickerMediaChosen(StoryEditorMedia.kt:99)
        at org.wordpress.android.ui.stories.StoryComposerActivity.handleMediaPickerIntentData(StoryComposerActivity.kt:367)
        at org.wordpress.android.ui.stories.StoryComposerActivity.onLoadFromIntent(StoryComposerActivity.kt:280)
        at com.wordpress.stories.compose.ComposeLoopFrameActivity.onCreate(ComposeLoopFrameActivity.kt:567)
        at org.wordpress.android.ui.stories.StoryComposerActivity.onCreate(StoryComposerActivity.kt:145)
        at android.app.Activity.performCreate(Activity.java:8000)
        at android.app.Activity.performCreate(Activity.java:7984)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3404)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3595) 
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85) 
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135) 
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066) 
        at android.os.Handler.dispatchMessage(Handler.java:106) 
        at android.app.ActivityThread.main(ActivityThread.java:7660) 
```

Steps:
1. Current develop as of time of writing (4746f7f224bda2502aac4e9fb284a5a28ad0a857)
2. Switch stories submodule to tip of this branch
3. Tab fab, select story option
4. Select 2 or 3 media items
5. Tap checkmark, observe crash

mzorz:
Thanks for reporting this @aforcier  - I think I came across the same issue and fixed it in https://github.com/Automattic/stories-android/pull/684/commits/92c9d542c79657afe1da756674a40c5d88f06682 in PR #684, will check and get back here 👍

mzorz:
Just reporting back - it's kind of related but not exactly the same 😅 - will get back here when I've a solution

mzorz:
Alright, fixed that crash in https://github.com/wordpress-mobile/WordPress-Android/pull/14627 @aforcier - thanks for noticing! 🙌 

Ready for another round, this PR can be reviewed along with that other WPAndroid PR. Noteworthy the first time you try you'll find the other issue https://github.com/Automattic/stories-android/issues/682 which is fixed separately in https://github.com/Automattic/stories-android/pull/684.

aforcier (approved):
1. Confirmed the crash fix with https://github.com/wordpress-mobile/WordPress-Android/pull/14627
2. Went through the steps and everything worked well for the original fix (I wasn't able to reproduce the original issue but didn't run into any problems after the fix either 👍 )

PR #689, opened 2021-05-21
Fix: using actualColor.background to change background

mzorz:
Originally spotted in #634 and fixed in a temporary branch as indicated in https://github.com/Automattic/stories-android/pull/634#issuecomment-845355859 (see section `Smoke test`), cherry picked the commit here to make background color selection for text work again on the color picker.

Before:
Screenshot_20210521-111109

After:
Screenshot_20210521-111443


aforcier (approved):
:shipit:

PR #690, opened 2021-05-21
Remove synthetics from PhotoEditor

mzorz:
This PR only removes the use of synthetics from PhotoEditor (last place where it was being used in the Stories library), and only implements [view binding](https://developer.android.com/topic/libraries/view-binding) in the `getLayout()`  method when new added views get created.

In the rest of places where a reference to a view was needed, I intentionally left it using `findViewById()` because it would otherwise mean a bigger refactor: when we create and when we re-create views we need a reference to them to be able to attach a gesture listener to it, so in this sense the PhotoEditor doesn't work like a regular Fragment or custom view in that it makes sense to keep a reference to binding and take it from there. Given Views in PhotoEditor are much more dynamic than the layout representation of a regular application screen, a migration to using View binding deserves some more, deeper thought.

To test:
- smoke test the app and add some views, verify these get added and are shown correctly
- WPAndroid: verify you can create and publish a Story, and also edit it once saved.

### DIFF
--- a/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
+++ b/photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
@@ -475,7 +475,7 @@ class PhotoEditor private constructor(builder: Builder) :
         var rootView: View? = null
         when (viewType) {
             ViewType.TEXT -> {
-                with (ViewPhotoEditorTextBinding.inflate(layoutInflater)) {
+                with(ViewPhotoEditorTextBinding.inflate(layoutInflater)) {
                     rootView = root
                     tvPhotoEditorText.gravity = Gravity.CENTER
                     if (mDefaultTextTypeface != null) {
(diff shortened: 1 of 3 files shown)


photoeditor/src/main/java/com/automattic/photoeditor/PhotoEditor.kt
```diff
@@ -343,7 +344,7 @@ class PhotoEditor private constructor(builder: Builder) :
 
             val multiTouchListenerInstance = getNewMultitouchListener(this) // newMultiTouchListener
             setGestureControlOnMultiTouchListener(this, ViewType.EMOJI, multiTouchListenerInstance)
-            touchableArea.setOnTouchListener(multiTouchListenerInstance)
+            findViewById<View>(R.id.touchableArea)?.setOnTouchListener(multiTouchListenerInstance)
```
aforcier, line 347: I thought we were generally able to eliminate `findViewById` by using view binding - can you explain why it's necessary here (and below)?

mzorz: Just for the record this is explained above in the description - basically it merits more time than we can prioritize right now so to avoid introducing new bugs we're keeping the PR simple. That said, we're going to close this one in favor of https://github.com/Automattic/stories-android/pull/695, more information regarding the rationale we utilized can be found there 👍

mzorz:
Closing this one in favor of https://github.com/Automattic/stories-android/pull/695

PR #695, opened 2021-05-25
Remove synthetics from PhotoEditor 

mzorz:
After a brief discussion with @aforcier we decided it was best to remove the last trace of synthetics from the PhotoEditor class (which is removed in [Kotlin 1.4.20](https://youtrack.jetbrains.com/issue/KT-42121)) without half-introducing view binding, leaving things in a better place for a proper refactor using [ `viewbinding` ](https://developer.android.com/topic/libraries/view-binding/migration)to be analyzed in the future.

WordPress Android uses Kotlin 1.4.10 so, once it gets bumped there the Stories library will likely get a bump in Kotlin version here (for which the [`kotlin-parcelize`](https://developer.android.com/kotlin/parcelize) plugin will have to be introduced as well when that time comes)

This PR then only removes synthetics and falls back to using good old `findViewById`.

To test:
1. open the demo app
2. create a slide or two
3. add some views
4. verify you can drag them around, resize, rotate, edit, etc.


aforcier (approved):
:shipit:

PR #688, opened 2021-05-21
Feature/view binding

mzorz:
Fix #628 

Feature branch PR - will hold all related work to implementing view binding, including #634 👍

PR #698, opened 2021-05-28
Simplify view binding

mzorz:
Follow up to #628, targets #688

After giving view binding a bit more thought and discussing different points of view, I arrived at the conclusion it'd be best if we kept the Stories library codebase aligned in terms of the usage as per in WPAndroid.

The core of the problem here is to provide an easy way in which Fragments / Activities can use binding without having to remember about dereferencing a class variable that holds beyond the Fragment/Activity lifecycle. Hence the simple to use, yet somewhat complexly crafted solution using lifecycle observers.

I wonder, why don't we get rid of the reference completely? we could do like so like it's done here: https://github.com/wordpress-mobile/WordPress-Android/pull/14596/files#diff-39225ce810fd1f96abbd79fe623a339e99c4bb826c38df330eeb18b36eee9a87R44-R47 

```
       with(FragmentLoginNoSitesEmptyViewBinding.bind(view)) {
            initContentViews()
            initViewModel(savedInstanceState)
        }
```

And then using Kotlin extension methods as shown there (i.e. `iniContentViews` and `initViewModel` in the example above)

Another example can be found here https://github.com/wordpress-mobile/WordPress-Android/blob/70ce7de8817020d29768074bb408688436320220/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt#L72-L103 

Citing @renanferrari here (ref `paqN3M-hV-p2#comment-510`):

_[...] we came to the conclusion that one of the main advantages of using ViewBindings is the fact that it gives us explicit nullability for our views, meaning we now have a much clearer idea of where exactly our views might be null and where they won’t. This is really good and can prevent a lot of unexpected headaches, while using a delegate would make this implicit again._

In the cases where we needed to hold the binding field, we would end up with something like this:

```
private var binding: FragmentMainBinding? = null
  
override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
    super.onViewCreated(view, savedInstanceState)
    with(MyFragmentBinding.bind(view)) {
        binding = this
        myFirstField.text = ...
        setupViews()
        setupObservers()
    }
}
 
private fun MyFragmentBinding.setupViews() {
    mySecondField.text = ... // Explicit nullability using the non-null 'this'.
}
 
private fun onSomeKindOfCallback() {
    binding?.myThirdField.text = ... // Explicit nullability using the nullable field.
}
  
override fun onDestroyView() {
    super.onDestroyView()
    binding = null
}
```

_There’s still some boilerplate as we need to set the field to `null` on `onDestroyView()` and we’re transforming binding-related functions into extension functions, but I believe this a fair price to pay for more null-safety._

I think we should be able to use it like that in Stories. And when that's not possible (i.e. we still need to keep a reference for other reasons), IMO it's still better to keep it explicit to avoid maintainability problems and / or being able to spot the potential issue in a quicker way.

#### To test
- smoke test the demo app
- smoke test with WPAndroid


stories/src/main/java/com/wordpress/stories/compose/story/StoryFrameSelectorFragment.kt
```diff
@@ -108,25 +109,25 @@ class StoryFrameSelectorFragment : Fragment(R.layout.fragment_story_frame_select
     }
 
     private fun updateContentUiState(contentState: StoryFrameListUiState) {
-        (binding.storyFramesView.adapter as StoryFrameSelectorAdapter).addAllItems(contentState.items)
+        (binding?.storyFramesView?.adapter as StoryFrameSelectorAdapter).addAllItems(contentState.items)
```
malinajirka, line 112: ❓ ⚠️ It seems this was discussed when this PR was created (it's explicitly mentioned in the PR description). But I'd still want to bring it up again to double check it's what we want to use.

I'd personally prefer to use "!!" for these cases - it's even[ recommended by the official docs](https://developer.android.com/topic/libraries/view-binding#fragments). Otherwise, we might have bugs in the app we won't know about unless someone manually contacts customer support. 

cc @aforcier @oguzkocer pinging you since Mario is afk. Wdyt?

oguzkocer: I am always in favor of fail-fast principle, so I agree with you on `!!` usage instead.

Although this line of code is quoted in the PR description, I don't see anything specific about `binding?` vs `binding!!` decision there. If there isn't a specific decision, I'd say go with whatever WPAndroid is doing for now and then open up a discussion if you disagree with it, so we can collectively make a decision. However, it looks like there is a mix of `binding?` and `binding!!` usage in WPAndroid, so... 🤷 not really sure what to do. I guess maybe we go with whatever @aforcier feels it'd be best for this project and then open up a general discussion to unify the approach?

malinajirka: >  I don't see anything specific about binding? vs binding!! decision there

I was referring to this line in the example: `binding?.myThirdField.text = ... // Explicit nullability using the nullable field.`
But I guess it still doesn't mean there was a decision made. Let's wait for Alex's imput.
Btw are you blocked by the synthetic accessors -> if you are, we can merge it and fix this "minor" issue later.

oguzkocer: It's a blocker for AGP update to 4.2.0, but I might be able to work around it with this branch. I also have other repositories to get through, so worst case, I'll switch to those for now.

malinajirka:
I reviewed the changes in this PR + I tested them in both the sample app and WPAndroid. All looks good to me (we might want to consider making changes in [this comment](https://github.com/Automattic/stories-android/pull/698#discussion_r649364186), but we can do that later)

stories/src/main/java/com/wordpress/stories/compose/story/StoryFrameSelectorFragment.kt
aforcier: No preference here, our main concern was to be as consistent with WPAndroid as possible. Since it sounds like WPAndroid itself is not internally consistent, deferring to the official docs in this case seems the reasonable thing 👍 

(It's possible Mario had some intention here I'm unaware of but I don't recall an explicit decision.)

malinajirka: To follow up on this discussion - I tried to make the proposed changes in [this PR](https://github.com/Automattic/stories-android/pull/704). However, it turned out to be a bit more complicated than I anticipated and I have a feeling that Mario made this decision on purpose. Considering that this is a minor issue and we already use the same approach through-out the codebase I decided to close the PR.

PR #702, opened 2021-06-10
Remove kotlin android extensions

malinajirka:
This PR is built on top of https://github.com/Automattic/stories-android/pull/688. It removes no longer necessary "kotlin-android-extensions" which has been preventing us from updating to a newer gradle version.

**This PR needs to be merged after https://github.com/Automattic/stories-android/pull/688 gets merged.**

This is still a draft since I haven't tested the changes in WPAndroid.

Changes in this PR
- Removes kotlin-android-extensions'
- Updates kotlin from 1.4.1 to 1.4.2 so `kotlin-parcelize` plugin is available ([release notes](https://github.com/JetBrains/kotlin/releases/tag/v1.4.20) - AFAICT there are no changes which would affect us - We might need to update other repos to this kotlin version, but I assume we'll need to do that anyway)
- Adds `kotlin-parcelize` plugin to keep support for @Parcelize annotation which used to be part of `kotlin-android-extensions` but now lives in a standalone plugin (`kotlin-parcelize`)
- Suppresses "ParcelCreator" lint error - it's a false positive. Classes which use @Parcelize don't need to [manually implement Creator](https://developer.android.com/kotlin/parcelize).

To test:
Smoke test the app

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         maven { url "https://jitpack.io" }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.0'
+        classpath 'com.android.tools.build:gradle:4.0.2'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
         classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlinVersion"
         //classpath 'com.automattic.android:fetchstyle:1.1'
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlinVersion = '1.4.10'
+    ext.kotlinVersion = '1.4.20'
     ext.serializationVersion = '1.0-M1-1.4.0-rc'
     repositories {
         google()
--- a/stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
@@ -9,9 +9,9 @@ import com.wordpress.stories.compose.story.StoryIndex
 import kotlinx.parcelize.Parcelize
 import kotlinx.serialization.Serializable
 
+@SuppressLint("ParcelCreator")
 class StorySaveEvents {
     @Parcelize
-    @SuppressLint("ParcelCreator")
     data class StorySaveResult(
         var storyIndex: StoryIndex = 0,
         val frameSaveResult: MutableList<FrameSaveResult> = mutableListOf(),
@@ -25,19 +25,16 @@ class StorySaveEvents {
         }
     }
     @Parcelize
-    @SuppressLint("ParcelCreator")
     data class FrameSaveResult(val frameIndex: FrameIndex, val resultReason: SaveResultReason) : Parcelable
 
     @Serializable
     sealed class SaveResultReason : Parcelable {
         @Serializable
         @Parcelize
-        @SuppressLint("ParcelCreator")
         object SaveSuccess : SaveResultReason()
 
         @Serializable
         @Parcelize
-        @SuppressLint("ParcelCreator")
         data class SaveError(
             var reason: String? = null
         ) : SaveResultReason()
--- a/stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
@@ -1,5 +1,6 @@
 package com.wordpress.stories.compose.frame
 
+import android.annotation.SuppressLint
 import android.os.Bundle
 import android.os.Parcelable
 import com.wordpress.stories.compose.frame.StorySaveEvents.SaveResultReason.SaveError
@@ -10,6 +11,7 @@ import kotlinx.serialization.Serializable
 
 class StorySaveEvents {
     @Parcelize
+    @SuppressLint("ParcelCreator")
     data class StorySaveResult(
         var storyIndex: StoryIndex = 0,
         val frameSaveResult: MutableList<FrameSaveResult> = mutableListOf(),
@@ -23,16 +25,19 @@ class StorySaveEvents {
         }
     }
     @Parcelize
+    @SuppressLint("ParcelCreator")
     data class FrameSaveResult(val frameIndex: FrameIndex, val resultReason: SaveResultReason) : Parcelable
 
     @Serializable
     sealed class SaveResultReason : Parcelable {
         @Serializable
         @Parcelize
+        @SuppressLint("ParcelCreator")
         object SaveSuccess : SaveResultReason()
 
         @Serializable
         @Parcelize
+        @SuppressLint("ParcelCreator")
         data class SaveError(
             var reason: String? = null
         ) : SaveResultReason()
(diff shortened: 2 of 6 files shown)


oguzkocer (approved):

malinajirka:
Merge of this PR is on hold until we migrate WPAndroid to kotlin 1.4.2. In order to do that we'll need to 
1. Replace synthetic accessors with ViewBinding in WPAndroid (probably even its dependencies)
2. Remove all usages of `apply plugin: 'kotlin-android-extensions'`

ParaskP7:
👋 @malinajirka !

> Merge of this PR is on hold until we migrate WPAndroid to kotlin 1.4.2.

Can you help me understand why is that? 🙏

build.gradle
```diff
@@ -1,15 +1,15 @@
...
+    ext.kotlinVersion = '1.4.20'
     ext.serializationVersion = '1.0-M1-1.4.0-rc'
     repositories {
         google()
         jcenter()
         maven { url "https://jitpack.io" }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.0'
+        classpath 'com.android.tools.build:gradle:4.0.2'
```
ParaskP7, line 12: Question (❓): Why was AGP updated as well?

stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
```diff
@@ -1,15 +1,17 @@
...
 import com.wordpress.stories.compose.frame.StorySaveEvents.SaveResultReason.SaveError
 import com.wordpress.stories.compose.frame.StorySaveEvents.SaveResultReason.SaveSuccess
 import com.wordpress.stories.compose.story.StoryIndex
-import kotlinx.android.parcel.Parcelize
+import kotlinx.parcelize.Parcelize
 import kotlinx.serialization.Serializable
 
 class StorySaveEvents {
     @Parcelize
+    @SuppressLint("ParcelCreator")
```
ParaskP7: Suggestion and Minor (💡 + 🔍 ): I would have just suppressed the whole `StorySaveEvents` class once instead of doing it multiple times, per inner class. Wdyt?

ParaskP7 (approved):

build.gradle
malinajirka: Good question, but I don't have a good answer :P. Someone mentioned that I should also update AGP on Slack so I just blindly updated it, I just checked the release notes.

stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
malinajirka: Good call, fixed in 8fa1f7c1c94be660127639490d84cce6ab1469ef.

I didn't realize I can suppress it at class level when the inner classes are not children of the top level class - but it make sense it works without issues.

malinajirka:
Thanks so much for checking out this PR @ParaskP7.

> Can you help me understand why is that? 🙏

'kotlin-android-extensions' and `kotlin-parcelize` can't live in the same project. Since stories are just a git submodule in WPAndroid, we need to remove all usages of 'kotlin-android-extensions' from WPAndroid before we'll be able to use Stories with `kotlin-parcelize`. At least that's how it looked when I tried to use this branch in WPAndroid - if I recall correctly the build actually worked, but AS was totally confused and some classes were underlined. Having said all that I didn't dive deeply into it as I knew we'd need to remove 'kotlin-android-extensions' in WPAndroid anyway.

build.gradle
oguzkocer: I don't remember if it was necessary for this PR, but it's something we want to do, so if there is a reason not to, let's keep it.

Otherwise I am just going to open a PR to update it :)

malinajirka: I think it wasn't necessary, but as you mentioned, it's something we want to do anyway.

ParaskP7: 👋 @malinajirka !

Thanks for the clarification. 🙏 

If both of you feel confident about this change and want it here, I have no objection with it as well. I was just trying to make sure I am not missing something. ✅

stories/src/main/java/com/wordpress/stories/compose/frame/StorySaveEvents.kt
ParaskP7: 👋 @malinajirka !

Thanks for the fix. 🙏

PS: I got confused by the mentioned commit at first, but then I saw this one: 8988aab5efb9bdd4d249144122936fce51b73f70 😅

ParaskP7:
👋 @malinajirka !

> 'kotlin-android-extensions' and kotlin-parcelize can't live in the same project. Since stories are just a git submodule in WPAndroid, we need to remove all usages of 'kotlin-android-extensions' from WPAndroid before we'll be able to use Stories with kotlin-parcelize.

🤔 I wouldn't think that to be the case, since in the end those are just story modules that will share their `.aar` with the downstream app. This is very interesting, thanks for the clarification. 🙏 

> At least that's how it looked when I tried to use this branch in WPAndroid - if I recall correctly the build actually worked, but AS was totally confused and some classes were underlined. Having said all that I didn't dive deeply into it as I knew we'd need to remove 'kotlin-android-extensions' in WPAndroid anyway.

🤔 This is very interesting, thanks for sharing. 🙏

As long as you have tested it and found this to be the case, I am fine to have this PR sitting as draft until we are done with the rest of `kotlin-android-extensions` in WPAndroid.

PS: @oguzkocer do we have any plans to also make the `stories-android` a composite build in the future? I kind of found it difficult to test this locally with WPAndroid as I didn't want to mess with submodules.

oguzkocer:
> I wouldn't think that to be the case, since in the end those are just story modules that will share their .aar with the downstream app. This is very interesting, thanks for the clarification.

@ParaskP7 I don't want to speak about this particular case without testing, but in our current setup, these modules are not any different from the `libs/abc` modules we have in WPAndroid. They are not standalone builds that are shared as an `aar` binary and does get affected by the configuration of the root project. Actually, if I am not mistaken, you can't even build some of the modules in isolation because they are not setup correctly for it.

> PS: @oguzkocer do we have any plans to also make the stories-android a composite build in the future? I kind of found it difficult to test this locally with WPAndroid as I didn't want to mess with submodules.

Yes, it'll be gone after Gradle 7 upgrade work is done. Having said that, it isn't any harder to use submodules than to use composite builds. You basically check out the branch in `libs/stories-android` folder of WPAndroid. The only issue is that sometimes git will hate you for using submodules and will tell you something doesn't exist even though it does. So... yeah, it'll be gone soon :)

P.S: I love hearing about composite build being a success. I've had my worries 😟

ParaskP7:
👋 @oguzkocer !

> @ParaskP7 I don't want to speak about this particular case without testing, but in our current setup, these modules are not any different from the libs/abc modules we have in WPAndroid. They are not standalone builds that are shared as an aar binary and does get affected by the configuration of the root project. Actually, if I am not mistaken, you can't even build some of the modules in isolation because they are not setup correctly for it.

Thanks for the clarification. 🙏 

> Yes, it'll be gone after Gradle 7 upgrade work is done. Having said that, it isn't any harder to use submodules than to use composite builds. You basically check out the branch in libs/stories-android folder of WPAndroid. The only issue is that sometimes git will hate you for using submodules and will tell you something doesn't exist even though it does. So... yeah, it'll be gone soon :)

`git will hate you for...` 😆 Thanks Oguz, yeap you are right, you can do that, thanks for sharing. It just scares me, I am not being able to trust git submodules lately, especially since I am not working regularly with them.

> P.S: I love hearing about composite build being a success. I've had my worries 😟

Yeap, big success and you are a 🦸 ! MANY THANK 🙇

ParaskP7:
👋 @malinajirka @oguzkocer !

This PR can be now effectively unblocked as all the work within the parent [Replace Synthetic Accessors with ViewBinding #14845](https://github.com/wordpress-mobile/WordPress-Android/pull/14843) issue for [WordPress-Android](https://github.com/wordpress-mobile/WordPress-Android) is done.

🍾 

PS: Let me know if I can help with anything here.

malinajirka:
I've rested this PR with the current `develop` of WPAndroid and I didn't encounter any issues - both the build and AS seemed to be happy with the current setup. I merging this PR, I'll create a PR in WPAndroid in a bit.

PR #703, opened 2021-06-11
Add view binding for IntroActivity in demo app

aforcier:
Fixes an issue where a fresh install of the Loop app would crash on launch in the view binding branch. It looks like the view binding setup for `IntroActivity` was missing, probably nobody noticed because the activity isn't shown once you've already granted the permissions the app needs.

### To test

1. Uninstall the `loop` app if you have it installed
2. Build this branch (there should be an installable build comment made automatically on this PR)
3. Confirm that you see an intro screen, you're able to grant permissions, and are taken to the next screen (which says "Still working on this! Tap the + to continue.")


app/src/main/java/com/automattic/loop/intro/IntroActivity.kt
```diff
@@ -9,13 +9,17 @@ import androidx.navigation.Navigation
...
 import com.automattic.loop.R
+import com.automattic.loop.databinding.IntroActivityBinding
 import com.automattic.photoeditor.util.PermissionUtils
 
 class IntroActivity : AppCompatActivity(), IntroFragment.OnFragmentInteractionListener,
     PermissionRequestFragment.OnFragmentInteractionListener {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val binding = IntroActivityBinding.inflate(layoutInflater)
```
malinajirka: 🔍 Nitpick: It seems we use
```with(IntroActivityBinding.inflate(layoutInflater)) {```
at most places instead of storing the binding into a local variable. Tbh I don't see much difference as long as it's a local variable and not a field. But I wanted to bring it to your attention in case you just missed it.

malinajirka (approved):
I'm not sure how I could missed this 😞. Nice catch! I've reviewed the changes and left a minor comment. Feel free to merge it no matter if you update the code or not.

app/src/main/java/com/automattic/loop/intro/IntroActivity.kt
aforcier: Oh sure - I emulated I think `PhotoPickerActivity` which does it the local variable way but also has slightly more involved setup. I like this pattern better, updated in 6d80f98.

malinajirka: 👍 Double checked and it works as expected:) Thanks

PR #704, opened 2021-06-11
Fail Fast - Replace usages of binding?. with binding!!. 

malinajirka:
This PR replaces usages of "binding?" with
- a non-nullable lateinit var in Activites
- "binding!!." in Fragments

This is a recommended practice by google and was discussed [here](https://github.com/Automattic/stories-android/pull/698#discussion_r649364186).

I also updated one place where we were setting `binding = null` in `onDestroyView` before `super.onDestroyView`.  This lead to a NPE as "dialog's onDismiss" was getting called when the binding was already null. (update: this introduced another issue, read the inline comment)

To Test:
1. Smoke test the app

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/text/TextEditorDialogFragment.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/text/TextEditorDialogFragment.kt
@@ -29,7 +29,9 @@ import com.wordpress.stories.databinding.AddTextDialogBinding
  */
 
 class TextEditorDialogFragment : DialogFragment() {
-    private var binding: AddTextDialogBinding? = null
+    private var _binding: AddTextDialogBinding? = null
+    // This property is only valid between onCreateView and onDestroyView.
+    private val binding get() = _binding!!
 
     @ColorInt private var colorCode: Int = 0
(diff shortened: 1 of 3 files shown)


stories/src/main/java/com/wordpress/stories/compose/text/TextEditorDialogFragment.kt
```diff
@@ -152,11 +154,9 @@ class TextEditorDialogFragment : DialogFragment() {
     }
 
     override fun onDismiss(dialog: DialogInterface) {
-        binding?.let {
-            val inputText = it.addTextEditText.text.toString()
-            textEditor?.onDone(inputText, TextStyler.from(it.addTextEditText, typefaceId, backgroundColorCode))
-            textEditorAnalyticsHandler?.report()
-        }
+        val inputText = binding.addTextEditText.text.toString()
```
malinajirka, line 157: ⚠️ This is now called even when the dialog is dismissed by the OS when it's destroying the activity -> we don't want to run this code in this scenario.

malinajirka:
I decided against introducing these changes as they are too risky without having a deeper knowledge of the codebase. We access the `binding` from public methods so it's a bit difficult to track down that they are getting called only between "onCreateView" and "onDestroyView".

Issue #499, opened 2020-08-15
Accessibility Improvements


aforcier:
Main issue to collect accessibility improvements notes.

* [ ] Look into `@SuppressLint("ClickableViewAccessibility")` annotations we currently have, resolve them
* [x] Add text to image/video description when uploading

mzorz:
> Look into @SuppressLint("ClickableViewAccessibility") annotations we currently have, resolve them

Documentation on how to handle accessibility with custom views found here https://developer.android.com/guide/topics/ui/accessibility/custom-views.html

PR #706, opened 2021-06-25
Address WPAndroid ClickableViewAccessibility lint errors

aforcier:
It looks like the view binding update (#688) created a few cases of the `ClickableViewAccessibility` lint check. Nothing changed feature-wise, I'm guessing the check just parses view binding based code differently and recognized something where it didn't before. I noticed this in https://github.com/wordpress-mobile/WordPress-Android/pull/14928, which is trying to pull the view binding changes from this library into WPAndroid.

`ClickableViewAccessibility` is a warning by default, but WPAndroid has promoted this to an error.

We have an open issue (#499) to eventually try to address some `ClickableViewAccessiblity` suppressions we already have, which I believe were imported with the original photoeditor library. (It may in fact be a true false positive in some cases due to how we're handling clicks.)

For now, to unblock that PR, I'm suppressing the lint check in the areas where it's now being flagged. I've also imported the `lint.xml` file from WPAndroid to make this less likely to happen in the future (I removed a few obvious WPAndroid-specific things). Ideally this will be part of better tooling in the future allowing us to share the `lint.xml` between projects with local custom rules.

### To test
* Check out ce3fe2f, run lint, confirm you get 11 errors (same as the WPAndroid PR)
* Check out the branch, run lint, confirm no errors
* The above two should also be verifiable by looking at CI, I'm pushing the PR early on purpose so CI runs lint on ce3fe2f
* Build this branch in WPAndroid, run `lintWordpressVanillaRelease`, make sure there are no errors

### DIFF
--- a/lint.xml
+++ b/lint.xml
@@ -1,6 +1,81 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <lint>
+    <!-- INFO -->
+    <issue id="IconMissingDensityFolder" severity="info" />
+
+    <!-- WARNING -->
+    <issue id="RtlSymmetry" severity="warning" />
+    <issue id="UseSparseArrays" severity="warning" />
+
+    <!-- ERROR -->
+    <issue id="SimpleDateFormat" severity="error" />
+    <issue id="RtlCompat" severity="error" />
+    <issue id="DefaultLocale" severity="error" />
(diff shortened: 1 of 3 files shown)


malinajirka:
Works as expected, thanks! 🙇

malinajirka (approved):

PR #709, opened 2021-07-27
Remove build config pre build task

oguzkocer:
Gradle 7.1.1 or Android Gradle Plugin 4.2.2 does not like the `createBuildConfigFieldsFromProperties` pre-build task we use to generate the build config values. I've [looked back on when we added this](https://github.com/Automattic/stories-android/pull/635) and here is what the goal was:

> The fact that Android doesn't wrap its configuration in a task, but instead has to have everything present during the script evaluation pass means that we need a bunch of early returns in order for ./gradlew applyConfiguration to have a chance to copy gradle.properties into position before errors are emitted. If you know of another solution I'm open to it 😅

We were throwing an error when we tried to load `gradle.properties` before `applyConfiguration` task had a chance to create it. I don't think this error is necessary. If we were to build the app without the `gradle.properties` file, it would fail because the build config fields would be missing. So, I think we should get rid of the extra check and the pre-build tasks and make this project consistent with others.

What do you think @jkmassel?

To test:
* If CI is green - we don't need to test anything

_P.S: It might help to check `Hide whitespace changes` PR setting since the whole code block is shifted._

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -104,27 +104,23 @@ dependencies {
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
 }
 
-tasks.register('createBuildConfigFieldsFromProperties') {
-    // Add properties named "loop.xxx" to our BuildConfig
-    android.buildTypes.all { buildType ->
-        def inputFile = file("${rootDir}/gradle.properties")
-        def properties = loadGradleProperties(inputFile)
-        properties.any { property ->
-        if (property.key.toLowerCase().startsWith("loop.")) {
-                buildType.buildConfigField "String", property.key.replace("loop.", "").replace(".", "_").toUpperCase(),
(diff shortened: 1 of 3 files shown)


jkmassel (approved):
So...I feel like this is the correct solution technically, but maybe not from a developer experience perspective? 

IMHO the benefit to this setup is that the developer doesn't need to figure out why a given variable isn't set – they get info about the root cause right away. I don't love this check, but it is helpful for that.

I don't necessarily have a solution, but just want to bring up that one potential downside.

oguzkocer:
I will post an RFC soon for dropping these custom values from `gradle.properties`. Let's try to address this issue as part of that.

_P.S: I _think_ we can check the tasks being run to still check if `gradle.properties` exists, but I don't want to do that in isolation in this library. Not yet anyway._

PR #711, opened 2021-08-04
Update minSdk to 24

oguzkocer:
After an internal discussion, we decided to upgrade minSdk to for WordPress-Android to 24 as we have less than 5% of our users on the earlier versions of Android.

Besides upgrading the `minSdk`, it removes obsolete checks for the earlier versions of sdk. I suggest turning on the `hide whitespaces` option to review these changes.

**To test:**
* I think we should smoke test the app and then smoke test the [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/15135)? I am not sure if this change can have any effect in the app to be honest.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ ext {
     compileSdkVersion = 28
     buildToolVersion = '28.0.3'
 
-    minSdkVersion = 21
+    minSdkVersion = 24
     targetSdkVersion = 28
 
     lifecycleVersion = '2.2.0'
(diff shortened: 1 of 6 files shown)


oguzkocer:
@cameronvoell @aforcier I've requested a review from you for this PR and then reverted that because it looks like there are still a couple things to fix. I'll request a review again once everything is in order. I am sorry for the premature ping!

aforcier (approved):
Changes make sense, nice cleanup 👍

Smoke tested the Loop demo app, no issues. Also smoke tested the WPAndroid PR from the associated PR (only checked the Stories flow), everything (media picker/camera/videos/fonts/publishing) checks out ✅ 

:shipit:

PR #713, opened 2021-09-02
Publish to s3

oguzkocer:
This PR adds support for publishing `stories`, `mp4compose` & `photoeditor` artifacts to s3 using our [publish-to-s3 Gradle plugin](https://github.com/Automattic/publish-to-s3-gradle-plugin). It also adds support for plugin DSL which is something we are adding across all our projects.

The PR follows the same approach as the [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2083) that added publish-to-s3 support. If you're unfamiliar with that PR, it's definitely worth having a look over the description to understand some of the decisions.

The publishing tasks are handled in Buildkite and depend on each other. `mp4compose` will be the first one to be published, because it has no dependencies. Then, `photoeditor` will be published which depends on `mp4compose` and then `stories` will be published which depends on both `mp4compose` & `photoeditor`. We use a conditional dependency for these projects so that during publishing we can replace the project dependency block with the binary dependency of the project that was published. This helps verify that the previous artifact was published correctly and the dependency graph has a binary dependency instead of a project block. (which wouldn't work when it's added to a client) More details about this in the [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2083) as previously mentioned.

Plugin DSL changes were straightforward, so there isn't much to explain there.

**To test**
* Smoke test the example app
* Follow the testing instructions in https://github.com/wordpress-mobile/WordPress-Android/pull/15301

Note that I am leaving this PR as a draft because https://github.com/wordpress-mobile/WordPress-Android/pull/15301 needs to be ready to merge before we can merge this in. I'd also like to get this PR reviewed by the library maintainers as well, just in case there was anything that was missed.


stories/build.gradle
```diff
@@ -108,3 +112,21 @@ android.buildTypes.all { buildType ->
...
+                groupId "com.automattic"
+                artifactId "stories"
+                // version is set by 'publish-to-s3' plugin
+            }
+        }
+   }
+}
+
+publishToS3Plugin {
+    mavenPublishGroupId "com.automattic"
```
mokagio, line 130: I have a personal curiosity question about the group id choice. The two dependencies use `com.automattic.stories` while this one uses `com.automattic`. The way I read this structure is that by putting the two dependencies in `.stories` group we make it clear that they are part of and required by the `stories` library. Is that right?

Do those projects function well in isolation? That is, would it beneficial to _someone_ in the community if we distributed them as part of `com.automattic`?

I realize this is a real minutia 😄 I'm just curious about the background of this setup. 👍

mokagio (approved):
Tested this both with the local demo app and [as part of the WordPress Android  integration PR](https://github.com/wordpress-mobile/WordPress-Android/pull/15301#pullrequestreview-763778270). All looking good 👍

stories/build.gradle
oguzkocer: That's a good question and something I've given a fair bit of thought before I setup fluxc.

Although these sub-modules can be used as a direct dependency, it's unlikely that they'd be. And this setup still allows it, it just needs to be added as a dependency as `com.automattic.stories.mp4compose` instead of the `com.automattic.mp4compose`.

For me the decision came down to how we'd want the organize things in our S3 repo and how to make it more maintainable. Keeping everything in the `org.wordpress` & `com.automattic` groups would have meant that these 2 folders would get crowded very quickly and it'd be hard to know which artifact belongs to which project. Using this setup allows us to group everything under the project folders such as `stories` & `fluxc` with only some minor inconveniences.

I like the current trade-offs, but I am happy to discuss it further if you'd like and re-visit the structure.

mokagio: Thanks for the extra details. Thoughtful as always. Your plan makes sense to me 👍

oguzkocer:
@mzorz @cameronvoell Do you mind having a quick look at the changes in this PR before we merge it in? I don't think the changes have impact on the library since they are all tooling oriented, but I wanted to make sure.

I am not sure who is working on stories-android these days (maybe there isn't any active development right now?), so if there is anyone who should review this, please feel free to ping them as well. I think @aforcier is AFK this week, that's why I didn't request a review from him.

Note that there will be an announcement post with details, so if you're not familiar with how the new setup will work, please don't worry. I think almost all Android developers are now familiar with the new system since it's now used in gb-mobile, FluxC, login & utils libraries, but 🤷

mzorz (approved):
Changes look good @oguzkocer ! 🎉 
I've followed the test instructions for both this PR and https://github.com/wordpress-mobile/WordPress-Android/pull/15301 and verified everything is working correctly 👍 

Nice change here! 🙌

oguzkocer:
Thank you for the reviews @mzorz 🙇

PR #714, opened 2021-09-09
Add StoryLoadStart and StoryLoadEnd events

mzorz:
This PR adds two events that are triggered from method `onLoadFromIntent()`:
- `StoryLoadStart` is triggered right away upon method start, so any interested parties listening can track when the ComposeLoopFrameActivity is about to load a Story coming from any of the entry points set through [Intent extras](https://github.com/Automattic/stories-android/wiki/2.-API#entry-points)
- `StoryLoadEnd` is triggered in all cases within `onLoadFromIntent()` in which a Story has been loaded

**Note**: given we were adding new events which were similar to `StorySaveEvents` but of a different nature, I first created a new package to keep both classes separately for cleanliness (see c4382d8), but turns out given we're serializing the slides along with their save results on the WPAndroid app this meant previously saved stories would not be able to get deserialized after this change. So, given this, I decided to add the class right along the same package where the StorySaveEvents where.

To test:
- See accompanying WPAndroid PR https://github.com/wordpress-mobile/WordPress-Android/pull/15310, or
- alternatively: observe the desired event and make sure it gets delivered, by using a breakpoint.

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
@@ -88,6 +88,8 @@ import com.wordpress.stories.compose.ScreenTouchBlockMode.BLOCK_TOUCH_MODE_PHOTO
 import com.wordpress.stories.compose.ScreenTouchBlockMode.BLOCK_TOUCH_MODE_PHOTO_EDITOR_READY
 import com.wordpress.stories.compose.emoji.EmojiPickerFragment
 import com.wordpress.stories.compose.emoji.EmojiPickerFragment.EmojiListener
+import com.wordpress.stories.compose.frame.StoryLoadEvents.StoryLoadEnd
+import com.wordpress.stories.compose.frame.StoryLoadEvents.StoryLoadStart
 import com.wordpress.stories.compose.frame.FrameIndex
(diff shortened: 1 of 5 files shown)


stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
```diff
@@ -747,18 +749,23 @@ abstract class ComposeLoopFrameActivity : AppCompatActivity(), OnStoryFrameSelec
     }
 
     protected open fun onLoadFromIntent(intent: Intent) {
+        EventBus.getDefault().post(StoryLoadStart(storyIndexToSelect))
```
jd-alexander, line 753: Nitpick: I noticed the `EventBus` calls below have a comment above about the dispatching taking place. Do we want to put one here for consistency?

jd-alexander (approved):
I verified that both `StoryLoadEvents` events are called and propagated as expected by utilizing breakpoints. LGTM 🚢

stories/src/main/java/com/wordpress/stories/compose/ComposeLoopFrameActivity.kt
mzorz: That's nice! Added in 34817a3c 👍

PR #717, opened 2021-12-10
Drop jetifier

malinajirka:
This PR is part of [an effort of dropping Jetifier from WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/15675).

Changes
- Updates EventBus to v3.3.1 which is migrated to AndroidX - v3.3.1 is already being used in WPAndroid so it's technically tested already
- Updates Utils to the latest version which is migrated to AndroidX
- Disables Jetifier in example gradle properties

To Test:
Make sure to disable Jetifier in your local gradle.properties => Smoke test the example app

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -97,7 +97,7 @@ dependencies {
     implementation project(path: ':stories')
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion"
 
-    implementation('org.wordpress:utils:1.26')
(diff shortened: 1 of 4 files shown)


oguzkocer (approved):
Since this is a blocker for https://github.com/wordpress-mobile/WordPress-Android/pull/15675, I figured I'd review this. The PR only contains version changes with utils being the only major-ish update. I think we are safe to merge this.

PR #736, opened 2023-06-30
[Build] Enable Non-Transitive Resources

ParaskP7:
This PR is a prerequisite for the `Gradle 8.1.1 & AGP 8.0.2 Upgrade for WPAndroid, WCAndroid & Related Libs` project.

Platform Request: `pdnsEh-13V-p2`
Project Thread: `paaHJt-57Z-p2`

-----

This PR enables non-transitive resources (`android.nonTransitiveRClass`) for the project.

FYI: This behavior becomes the default in AGP `8.0` and higher. As such, this becomes a prerequisite for the AGP `8.0.2` upgrade, that is of course, unless `android.nonTransitiveRClass` is explicitly set to `false`.

-----

## To test:

1. Verify that all the CI checks are successful.
2. Smoke test the app.

-----

Please also notice this 2c297bf80e7da21d9ad1104f09ae89cd24a0fe0d and its description. This commit doesn't updates the encrypted `gradle.properties.enc` file as it seems that this kind of encryption related functionality for this project isn't supported anymore (could not locate `Gemfile` and all, but it seems was supported in the past).

Can you please confirm the above? 🤔 

And if that's the case, let's take this opportunity to remove any such encryption related functionality from this repo, plus everywhere else (ie the `android/Loop/gradle.properties` file).


oguzkocer (approved):
> Please also notice this https://github.com/Automattic/stories-android/commit/2c297bf80e7da21d9ad1104f09ae89cd24a0fe0d and its description. This commit doesn't updates the encrypted gradle.properties.enc file as it seems that this kind of encryption related functionality for this project isn't supported anymore (could not locate Gemfile and all, but it seems was supported in the past).

@ParaskP7 This repo uses the [configure Gradle plugin](https://github.com/Automattic/configure/) instead of the Ruby toolchain since we don't need `fastlane` for this project. However, the plugin version is outdated and doesn't work with Apple Silicon machines. I created a PR #737 to address that and to update the secrets hash.

I'll leave merging that PR to you since it's targeting this PR.

ParaskP7:
👋 @oguzkocer !

> @ParaskP7 This repo uses the [configure Gradle plugin](https://github.com/Automattic/configure/) instead of the Ruby toolchain since we don't need fastlane for this project.

Got it! 👍 

> However, the plugin version is outdated and doesn't work with Apple Silicon machines. I created a PR https://github.com/Automattic/stories-android/pull/737 to address that and to update the secrets hash.

Awesome, thanks for doing that! 🥇 

> I'll leave merging that PR to you since it's targeting this PR.

I am on it, thanks! 👀

ParaskP7:
Ready to merge, thank you for reviewing and testing this @oguzkocer , you rock! 🙇 ❤️ 🚀

PR #737, opened 2023-07-05
Update configure plugin to 0.6.5

oguzkocer:
Updates configure plugin to `0.6.5` and the `.configure` hash to latest to include change to the secret `gradle.properties` file that enables non transitive resources.

**To Test**

* Make sure to delete the `vendor` folder first. There is still a bug in the configure plugin where it doesn't update the binary if it's already downloaded.
* Run `./gradlew applyConfiguration`
* Verify that `android.nonTransitiveRClass=true` is included in the new `gradle.properties` file.


ParaskP7 (approved):
Thanks for this update @oguzkocer ! 💯 

FYI: I took the opportunity to further update the `pinned_hash` to that version of the encrypted `gradle.properties` file that points to `android.enableJetifier=false` as I noticed that to still point to a `true` value, which we need to avoid at this point.

Please take a look and merge this PR if you too feel that everything is okay with my change. 🙇 

PS: I actually had this `android.enableJetifier=false` change committed in secrets already few days ago, but silly me, I forgot to push it... 🤷

ParaskP7:
Thank you for merging this @oguzkocer , you rock! 🙇 ❤️ 🚀

PR #739, opened 2023-07-28
Add a guard to avoid calling Glide load when activity is destroyed

mkevins:
#### Related PR

https://github.com/wordpress-mobile/WordPress-Android/pull/18850

### Description

This PR adds a guard to avoid calling `Glide`'s `load` when the activity is destroyed, which otherwise leads to a crash on a failed assertion. More details can be found here: https://github.com/wordpress-mobile/WordPress-Android/issues/18809.

To test:

Test via the steps in [the description of the related PR](https://github.com/wordpress-mobile/WordPress-Android/pull/18850#issue-1825696147).


photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
```diff
@@ -190,7 +191,7 @@ class PhotoEditorView : RelativeLayout {
 
         backgroundImage.setOnImageChangedListener(object : BackgroundImageView.OnImageChangedListener {
             override fun onBitmapLoaded(sourceBitmap: Bitmap?) {
-                if (attachedToWindow) {
+                if (attachedToWindow && (context as? Activity)?.isDestroyed != true) {
```
irfano: To be extra safe, we can also check `!Activity.isFinishing`. Wdyt?

photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
```diff
@@ -190,7 +191,7 @@ class PhotoEditorView : RelativeLayout {
 
         backgroundImage.setOnImageChangedListener(object : BackgroundImageView.OnImageChangedListener {
             override fun onBitmapLoaded(sourceBitmap: Bitmap?) {
-                if (attachedToWindow) {
```
irfano, line 193: `attachedToWindow` is being set after ` super.onDetachedFromWindow()`. Actually, `View` already has `isAttachedToWindow` property. So, I think we can also use `View.isAttachedToWindow` in addition to our custom `attachedToWindow`.

irfano (commented):
Your analysis of the crash makes sense to me. Even though I couldn't reproduce the crash, this could potentially prevent it. I've added some comments and am waiting for you to consider them before approving and merging this PR.

photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
mkevins: Good idea! Resolved w/ https://github.com/Automattic/stories-android/pull/739/commits/34ebfa03006d2865fb5dfb7e0bccd352fdb22633

photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
mkevins: Resolved in https://github.com/Automattic/stories-android/pull/739/commits/bf7a3e19ec765fc126710b2f8212015d01e4cde5 (also I added a comment referencing this discussion, since the `if (attachedToWindow && isAttachedToWIndow ...` part of the code surely looks strange 😅).

Probably, at some point this should be refactored, but this is out of scope for this crash fix, imo. 🤷‍♂️

mkevins:
Thanks for reviewing this Irfan! This is ready for another pass.

photoeditor/src/main/java/com/automattic/photoeditor/views/PhotoEditorView.kt
irfano: > the if (attachedToWindow && isAttachedToWIndow ... part of the code surely looks strange 😅

I agree. 😄 Thanks for adding a comment.

> Probably, at some point this should be refactored, but this is out of scope for this crash fix, imo. 🤷‍♂️

I think the same.

irfano (approved):
Thank you for the improvement commits after my review. LGTM! 👍🏻

mkevins:
Thanks for reviewing and testing Irfan!

PR #741, opened 2023-09-12
Fix EmojiPicker crash when getting LayoutParams

thomashorta:
This crash was happening when flinging upwards in the Stories Compose screen, which should open the Emoji picker, because it was trying to get a Coordinator LayoutParams from a RelativeLayout (as it was using the wrong view).

Testing steps:
1. Install the demo app
2. Select / record a video
3. Swipe up from the bottom of the screen
4. **Verify** app doesn't crash and emojis are shown

### DIFF
--- a/stories/src/main/java/com/wordpress/stories/compose/emoji/EmojiPickerFragment.kt
+++ b/stories/src/main/java/com/wordpress/stories/compose/emoji/EmojiPickerFragment.kt
@@ -78,9 +78,8 @@ class EmojiPickerFragment : BottomSheetDialogFragment() {
         super.setupDialog(dialog, style)
         activity?.let { activity ->
             val binding = FragmentBottomStickerEmojiDialogBinding.inflate(LayoutInflater.from(context))
-//            val contentView = View.inflate(context, R.layout.fragment_bottom_sticker_emoji_dialog, null)
             dialog.setContentView(binding.root)
-            val params = binding.root.layoutParams as CoordinatorLayout.LayoutParams
+            val params = (binding.root.parent as View).layoutParams as CoordinatorLayout.LayoutParams
 
             @Suppress("DEPRECATION")
             (params.behavior as? BottomSheetBehavior)?.setBottomSheetCallback(bottomSheetBehaviorCallback)


ParaskP7 (approved):
👋 @thomashorta !

I have reviewed and tested this PR as per the instructions, everything works as expected, nice fix! 🌟

-----

`EXTRAS`
1. Warning (⚠️): This is not related this this fix, but rather this `emoji` swiping feature overall. Initially, I couldn't replicate the crash, just to then verify that the fix actually works. Me not being able to replicate the crash, meant that I wasn't even able to replicate the `emoji` feature. I then realized that this might be because of me using the `Gesture navigation` system, that is, instead of using the `3-button navigation` system. Thus, after me changing to the `3-button navigation` system, I was able to replicate the crash, and then its fix. This made me think:
    1. FWIW, I think we should deal with this issue at some point. I am sure we have lots of user using the `Gesture navigation` system nowadays. Wdyt? 🤔
    2. Also, [this discussion](https://github.com/wordpress-mobile/WordPress-Android/issues/17565#issuecomment-1403805689) I had with @mzorz on emoji support now makes more sense to me as it seems I couldn't test this feature back then... 🤷

thomashorta:
> FWIW, I think we should deal with this issue at some point. I am sure we have lots of users using the Gesture navigation system nowadays. Wdyt? 🤔

@ParaskP7  I was able to reproduce it even with Gesture Navigation on, it's just trickier because your finger has to be close enough to the bottom to satisfy our `onFling` conditions but not enough that it triggers the system gestures. 😅 

Honestly, I don't think this emoji feature was supposed to be available to the user as we don't really instruct them about it. It seems to me this was a left-over code from some point in the past and I simply wanted to make sure it at least doesn't crash, especially because it is actually active in Jetpack `stories` feature.

I checked Sentry for the past 90 days and only found [5 events of that crash](https://a8c.sentry.io/issues/4271652063/) in Jetpack, which I believe were users probably trying to do the System Gesture and getting the Emoji picker instead.

I'm merging the fix for now, but I wonder if this should just be turned off since it doesn't seem to be something we actually intended to be visible to users (based on the fact that there's no visible button or instruction regarding the gesture).

ParaskP7:
👋 @thomashorta and I agree with everything you said above, thanks for the reply! 💯 

> @ParaskP7 I was able to reproduce it even with Gesture Navigation on, it's just trickier because your finger has to be close enough to the bottom to satisfy our onFling conditions but not enough that it triggers the system gestures. 😅

For the life of me, I just couldn't reproduce it, my fingers seems to be too fat... 🤣 

> ...I wonder if this should just be turned off since it doesn't seem to be something we actually intended to be visible to users...

Personally, I am voting in favor of that, so yes! 👍

Issue #744, opened 2024-01-10
Upgrade stories-android Target SDK to 34


notandyvee:
- [ ] Update the `targetSDK` to 34

PR #745, opened 2024-01-19
[Fix] Update targetSdk to 34 and fixed a few issues along the way.

notandyvee:
Issue: https://github.com/Automattic/stories-android/issues/744
Parent: https://github.com/orgs/wordpress-mobile/projects/223

This PR updates the `targetSdk` version to 34. During testing, two issues were found:

1. Services now **require** a `foregroundServiceType`.
2. Sentry needed to be updated. Found an [issue](https://github.com/getsentry/sentry-java/issues/2986) on Sentry's repo and the [corresponding fix](https://github.com/getsentry/sentry-java/pull/2990). It was [released in 6.32.0](https://github.com/getsentry/sentry-java/releases/tag/6.32.0). I realize this increases risk, so I only updated Sentry to the lowest possible version that avoided the crash.

Something to note. I used [datasync](https://developer.android.com/about/versions/14/changes/fgs-types-required#data-sync) as the foreground service type. This makes sense as we are processing the image and the camera does not need to be open for this. But feel free to make sure I chose the right foreground service type as a second pair of eyes.

Note:

I tested this with the sample app and Wordpress Android. Worked fine on both.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,7 +86,7 @@ subprojects {
 ext {
     minSdkVersion = 24
     compileSdkVersion = 34
-    targetSdkVersion = 33
+    targetSdkVersion = 34
 }
 
 ext {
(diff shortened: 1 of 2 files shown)


irfano (approved):
`dataSync` looks proper for `FrameSaveService` to me.

> I only updated Sentry to the lowest possible version that avoided the crash.

Good call. There are still many versions between Sentry `5.01` and `6.32.0`; yes, there is a risk. I checked [changes between two versions](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md), and I couldn't find any change that may affect us.
So, LGTM! 👍🏻

